PR #12, opened 2017-07-23
adapt cycle by cycle amplitude estimate for possible slow hilbert transform

srcole:
scipy.signal.hilbert takes a long time for some seemingly arbitrary signal lengths. A way around this is to specify the number of points in the signal. This was not initially supported in the `shape` module, like it is in the `timefrequency` module.


srcole:
@TomDonoghue can you change the configuration for pep8speaks so that it ignores line length (

Issue #2, opened 2017-06-21
Burst detection


srcole:
Add a module for burst detection. The main function should take as input a signal and frequency range of interest, and output the samples at which the oscillation is present

AndrewJWashington:
Relevant PR: https://github.com/voytekresearch/neurodsp/pull/34

PR #17, opened 2017-08-29
added spectral domain functions in spectral.py, attempt #2

rdgao:
fixed a failing test in spectral.psd: spectrogram was not handling a None parameter


TomDonoghue:
PEP8 OR GO HOME!

neurodsp/tests/test_spectral.py
```diff
@@ -0,0 +1,101 @@
+"""
+test_timefreq.py
+Test functions in the time-frequency analysis module
+"""
+
+import pytest
```
TomDonoghue: You don't need to import pytest for a test file.

srcole: True. This should be changed for all test files.

neurodsp/tests/test_spectral.py
```diff
@@ -0,0 +1,101 @@
+"""
+test_timefreq.py
```
srcole: update the header for your test file

PR #27, opened 2017-09-03
added slope fitting function for PSD and tutorial for spectral toolbox

rdgao:


rdgao:
uhhh...scikit-learn 0.19.0 requires py3.6, while neurodsp supports py3.4. Help?

rdgao:
wait, no:
image

TomDonoghue:
I shoulda paid more attention to neurodsp: I also have some python slope fitting stuff (SfN poster related) in a repo I just invited you to (Richard). 

Not that related to this, mine was mostly exploring using data driven oscillation zones, which has been collapsed into FOOOF now (and so my repo will be superseded by FOOOF stuff), and this is same / better than slope fitting I had - just bring it up for code collection / auditing purposes.

TomDonoghue:
Version conflict: it looks like there is no sklearn 0.19 for py34 (there is sklearn for py27). 

Seems sorta weird - but it means if we really need to use sklearn >= 0.19, I think we need to drop support for py34. 

That would seem fine to me, but I don't actually know how to update Travis for that - probably Scott does.

srcole:
I think Tom's last suggestion is good. I think it makes more sense to support python 3.4 for now (I'm running 0.18.2 on python 3.6).

require sklearn v 0.18.1.

srcole:
Also: where is pep8speaks?

srcole:
Also. Add the new module and link to the tutorial to the list in the README.

srcole:
your tutorial is really good!

rdgao:
oh okay, sounds good. Thanks
Have fun in Seattle!
rg.

PR #28, opened 2017-10-17
updated scv outlier to only throw out high power windows

rdgao:


rdgao:
is this a problem? seems like a weird error that python 3.4-3.6 does not exist. or can I just merge it?
@srcole @TomDonoghue

TomDonoghue:
Not sure what happened with py34, but in 35 & 36 it looks like the test_fitpsd test failed, so that seems like an issue?

Issue #37, opened 2018-01-12
Error when trying to increase signal duration for hilbert transform


srcole:
There's a phenomena that the hilbert transform function (scipy.signal.hilbert) runs very slowly for some durations of signals. This can be bypassed by specifying the number of points in the FFT. I've noticed that if I make the number of points equal to the next power of 2, that helps.

However, the current code to implement this when calculating an amplitude envelope (`_hilbert_ignore_nan` in timefrequency.py) forgets to decrease this signal back to the original length.

Issue #36, opened 2017-12-08
PYPI stuff


TomDonoghue:
I think we can drop the egg-info file on master branch. AFAIK it's not needed - it's actually out of date (it's the info for v0.1 release), so that actually makes it sort of confusing to have hanging around, with the current tagged version (also on PYPI) being 0.2.

Also: we should probably add a description to show up on PYPI:
https://pypi.org/project/neurodsp/

srcole:
PR #38 dropped the files.

As for the PYPI description, we should probably translate the readme into [restructured text](http://www.sphinx-doc.org/en/stable/rest.html). [MNE](https://pypi.python.org/pypi/mne/0.15) has a nice example in the mne.egg-info/PKG-INFO file

TomDonoghue:
Yeah... going for RST is probably the ultimately better option. I didn't bother doing that with FOOOF, and so just copied in a snippet of the README into setup.py, which is also a (somewhat hacky) option.

TomDonoghue:
This is done right? We can close it?

PR #38, opened 2018-01-12
Fix hilbert issue #37

srcole:


srcole:
Note this also now removes the egg-info folder as in issue #36

Issue #46, opened 2018-02-23
New module: simulating oscillatory neural signals


srcole:
As discussed with @rdgao.

We should have a function (or set of functions) that allow us to construct signals that contain oscillations with various properties (e.g. stationary or bursty) along with noise (probably brown noise), and control the relative power between these two.

Some current code snippets below can be used as a basis.

Simulate a stationary oscillator with arbitrary rise-decay symmetry:
```
import numpy as np
from scipy import signal

rdsym = .4
period_ms = 100
Fs = 1000
N_cycles = 50

rise_ms = np.round(period_ms * rdsym)
decay_ms = period_ms - rise_ms

pha_one_cycle = np.hstack([np.linspace(0, np.pi, decay_ms+1), np.linspace(-np.pi, 0, rise_ms+1)[1:-1]])
phase_t = np.tile(pha_one_cycle, N_cycles)
x_sawtooth = np.cos(phase_t)
t = np.arange(0, len(x_sawtooth)/Fs, 1/Fs)
```

High-pass-filtered brown noise:
```

def simfiltonef(T, f_range, Fs, N):
    """ Simulate a band-pass filtered signal with 1/f^2 
    Input suggestions: f_range=(2,None), Fs=1000, N=1000
    
    Parameters
    ----------
    T : float
        length of time of simulated oscillation
    Fs : float
        oscillation sampling rate
    f_range : 2-element array (lo,hi)
        frequency range of simulated data
        if None: do not filter
    N : int
        order of filter
    """

    if f_range is None:
        # Do not filter
        # Generate 1/f^2 noise
        brownN = simbrown(int(T*Fs))
        return brownN
    elif f_range[1] is None:
        # Make filter order odd if necessary
        nyq = Fs / 2.
        if N % 2 == 0:
            print('NOTE: Increased high-pass filter order by 1 in order to be odd')
            N += 1
            
        # Generate 1/f^2 noise
        brownN = simbrown(int(T*Fs+N*2))

        # High pass filter
        taps = signal.firwin(N, f_range[0] / nyq, pass_zero=False)
        brownNf = signal.filtfilt(taps, [1], brownN)
        return brownNf[N:-N]

    else:
        # Bandpass filter
        # Generate 1/f^2 noise
        brownN = simbrown(int(T*Fs+N*2))
        # Filter
        nyq = Fs / 2.
        taps = signal.firwin(N, np.array(f_range) / nyq, pass_zero=False)
[...]

PR #54, opened 2018-03-29
add bursting to simulation module

srcole:
More fully addresses issue #46

Addition to the last PR #53 that adding simulating oscillators with noise. Now the module can simulate bursting oscillations in addition to stationary ones.


srcole:
pep8speaks never yelled at me, so I guess there are no pep8 issues? or it's broken.

PR #58, opened 2018-04-11
added exponential synapse noise and jittered oscillator simulation

rdgao:


srcole:
looks good!
Thanks for adding.
just that 1 pep8 issue.

PR #66, opened 2018-06-29
Updates to setup, etc. 

TomDonoghue:
Hey y'all, 

First off: sorry, I made a bit of a mess at first, by accidentally pushing changes straight to the voytekresearch repo instead of my fork (cloned the wrong one... ooops). 

Anyways - main repo has been stepped back to where it was. 

Suggested changes:
- move version number inside the module, so it is accessible from the code (including a change to setup.py to read version number from the module, so that version is only specified in one place). 
- update setup.py: add more classifiers, and update install_requires
- update README: add a couple more badges, and update the install instructions. 

Note: don't merge yet, I think there is a bit of work to sort out dependencies, and I'm not 100% sure my setup.py is currently totally proper.


srcole:
Sounds great!

TomDonoghue:
Open questions about dependencies:
- Is there a reason the requirements.txt file in the repo has very specific versions set? If not really, can we relax those specifiers?
- Are there minimum requirements for dependencies? (something like scipy >= 0.17 (or similar) is pretty common, since something pretty big changed then, I think).

srcole:
requirements.txt: I saw Erik do it that way before, so that's what I adopted, but agree that a minimum would be more ideal. But I wouldn't know what the minimum version is. Like you said: Maybe scipy >=0.17?

srcole:
looks good. thanks for adding these things! It good to merge?

TomDonoghue:
I was going to look into dependencies a bit more to figure out best practices there, and maybe recommend a change to requirements.txt, but if I do get around to that, it can just go into a different update. 

So yeh - if everything looks good to you, then I think this should be ready to merge!

srcole:
oh right, I forgot about the dependencies. Yeah. Later.

TomDonoghue:
I don't understand why, because the tests passed pre-merge, but now Travis is failing, only on 3.6 - with exit code 2 - which apparently reflects 'test execution interrupted by user' (?!?)....

TomDonoghue:
@srcole re above failing tests: when I run locally, I do pass, but get some warnings that appear only on 3.6 (not older versions). Do you have any idea what's going on? I don't understand how this update could have changed something to make Travis fail only on 3.6, in a way that doesn't happen locally....

Issue #67, opened 2018-07-10
sim suggestions


TomDonoghue:
I'm starting to use neurodsp.sim, and have some possible suggestions:

ToDos:
- [x] Add in general colored noise generation (implementation available here: https://github.com/felixpatzelt/colorednoise) [Note: Richard added his own implementation].
- [x] The 'Input Suggestion' docs for sim_oscillator or wrong / misplaced, right? Am I missing something? [Fixed: with more general doc clean ups.]
- [x] I'll probably update some variable names, to update to our API conventions (ex - brownNf -> brown_nf or similar). 

Open Questions:
- Is there any reason not to generalize all the functions that currently add 1/f^2 noise, to be able to add 1/f^n noise, with set-able n?
- Mixed conventions: some functions take in a signal length, others a number of samples. We should consolidate on one approach - any suggestions on which is better?

Anyways - I'll use this issue for continuing points of discussion, and open a PR soon-ish with some updates - so let me know of any thoughts about things here.

rdgao:
- I have code for generating arbitrary exponent noise through a rotation, seems to be how that colorednoise repo does it. I can add it to sim.py if you want: https://github.com/rdgao/SpaceRecon/blob/master/nlds/utils.py

- I think there are currently two ways to get 1/f^2 noise in sim.py: my prefered way has been convolving with an exponential (synaptic) kernel, I think Scott's been using integrated white noise with a highpass. No real reason not to have a third way, as above, I guess there hasn't been a need to vary slope? In any case, yeah I can add that.

- yes please update variable names as per the convention. For the signal length thing, the most "friendly" user-level function, in my experience, is to take in a length (in seconds) and fs, so everything that uses real time is scaled appropriately. But under the hood, I think it almost always just converts that to a signal length in samples. I would prefer the former as far as user-facing functions go.

srcole:
> The 'Input Suggestion' docs for sim_oscillator or wrong / misplaced, right? Am I missing something?

Yeah, you're right. That was an earlier draft of inputs

Also agree that the API should be improved. My bad :)

TomDonoghue:
@rdgao : If you have code for generating 1/f^x noise, that you can add in, that would be great! If it's the same as in the repo I linked great (if it's a different procedure, maybe we should check equivalence, explore a bit?). If you want to make a PR with that copied in, I'll start working with that updated version, and go from there with the other things I was thinking of.

TomDonoghue:
[...]

PR #68, opened 2018-07-13
added variable 1/f exponent simulation

rdgao:
#67


rdgao:
@srcole @TomDonoghue failing for 3.6 here, but not on my machine. You guys had this problem very recently right?

srcole:
right. I think we should just undo some of the setup.py changes in PR#66 until it doesn't fail anymore. and then re-investigate later

rdgao:
okay, sounds good. So should I just merge here?

srcole:
i think it makes sense for me to make a quick PR to undo the setup.py stuff and merge it. and then you can get that update in your PR and we can make sure that travis is OK with everything you changed

rdgao:
oui tres bien

Issue #70, opened 2018-09-25
Remove waveform shape and PAC functions


srcole:
Waveform shape analysis should be done using bycycle: https://github.com/voytekresearch/bycycle

PAC should be done using pacpy: https://github.com/voytekresearch/pacpy
or pactools: https://github.com/pactools/pactools

PR #71, opened 2018-09-25
Remove duplicate PAC and shape functions

srcole:
Issue #70


neurodsp/version.py
```diff
@@ -1 +1 @@
-__version__ = '0.3.1'
\ No newline at end of file
+__version__ = '1.0.0'
```
rdgao: are we there? there's more simulation stuff to be added right?

srcole: No. But I thought the convention was to have the repo have the version of the next release. is that right @TomDonoghue ?

TomDonoghue: Yeh, I think that's standard (it's what I do) is that the Github has the next version number, while it's being worked on. Perhaps slightly better is to have a dev tag, so like:
pip version is (for example): 0.3.4
GitHub working version: 0.4.0-dev
^Then remove the '-dev' when tagged & released. 

Main thing is if the version on Github has things beyond the tagged release, it should have a different / higher version number, so this covers that.

Issue #61, opened 2018-06-01
f_c float/tuple option is really ugly


rdgao:
I think having to juggle between float and tuple is confusing, as oppose to always requiring a tuple and put in 0s for low/high-pass.

srcole:
agreed. but can you do it? :)

rdgao:
ffs.

srcole:
I think it's reasonable to always accept tuples AND also accept an int/float if the filter type is highpass or lowpass

srcole:
@rdgao What do you think needs to be changed for the filtering to be ready for v1.0?

rdgao:
bandpass/bandstop requires tuple, and tuple and float are acceptable for low/high-pass? 

so lowpass would only look at fc[1], and highpass fc[0]?

srcole:
sounds good

rdgao:
so am I doing it?

srcole:
looks like you're assigned to this issue

rdgao:
should we deprecate the option to use f_lo/f_hi?

rdgao:
alright I'm taking it out. I'm the captain now.

rdgao:
solved in PR#72

PR #72, opened 2018-09-29
Fixed fc argument in filter, and added PSD rotation function

rdgao:


rdgao:
#61 and https://github.com/voytekresearch/fooof/issues/98

PR #73, opened 2018-10-01
Update simulation functions (mostly docs & style)

TomDonoghue:
Edit: updated to reflect actual changes:

This PR updates a bunch of the simulation functions, mostly in terms of updating docs & style. The edits here should be mergeable. This is in partial fulfillment of #67. In a later PR, I will finish some things from #67, in terms of generalizing some of these functions - but that can be independent of this PR.


rdgao:
Isn't variable exponent already in sim.py? 

I actually just changed it, grabbed the rotation code out of there and put it in spectral.py, only to realize that request was on fooof, not neurodsp lol.

TomDonoghue:
Variable exponent for time-series, not PSDs (see #67).

But yeh - the PSD rotation thing should be in FOOOF.

rdgao:
wait I'm confused. So I added the arbitrary exponent **time series** sim was added here some time ago, and you want the arbitray exponent **spectrum** generation to be in fooof right?

The spectral.rotate_powerlaw() will take a spectrum and rotate it, so you can use this to generate whatever exponent you want. I guess I can make a PR on fooof to get my t-shirt.

PR #74, opened 2018-10-01
filter param, power law time series simulation & removed fit_slope

rdgao:
- filter.py: allow tuple for filter param when lowpass or highpass, removed transition band warning and print out bandwidth in verbose mode instead
- sim.py: restructured sim_variable_powerlaw(), moved rotate_powerlaw() to spectral.py
- spectral.py: removed fit_slope() due to scope overlap with FOOOF, and moved to be a private function in burst.py


srcole:
looks good to merge

PR #75, opened 2018-10-09
add arbitrary power law function to init.py

rdgao:


srcole:
is this just to get another PR for hacktoberfest....

TomDonoghue:
Do we really want every sub-function aliased through the init?

srcole:
I initially did this cause I thought it was kind of nice to just have:

```
import neurodsp
f, psd = neurodsp.psd(x, Fs)
```

as opposed to something like:
```
from neurodsp import spectral
f, psd = spectral.psd(x, Fs)
```

rdgao:
yeah, it's a nice plus so I don't have to think about the submodule. 

But practically, I'd have to look up what the functions I want to use are anyway, so I'm not married to it. I'm just following convention in this PR

srcole:
yeah, since it's following convention, I'm gonna merge it now, but if @TomDonoghue or someone else things we should change it, feel free to open up an issue!

TomDonoghue:
I think it's a scale thing - for now, I think it's good / fine, but as things grow having everything in the base name space could become a bit of a mess, and can end up making things harder to find (imagine if scipy aliased everything to base namespace). 

Of course, we are not massive, but if there could be growth, then it's maybe easiest to change the convention for our big breaking release, rather than have to backtrack much later.

PR #78, opened 2018-10-23
truncated and outlier removed spectral.psd, value checking in spectral.rotate_powerlaw

rdgao:
in spectral.psd():
- added flim option to return truncated psd
- added spg_outlierpct option to through out a percentage of spectrogram slices prior to averaging.

in spectral.rotate_powerlaw():
- check for whether f_axis[0] is 0Hz before defaulting to unity
- check that f_rotation is in range of f_axis

### DIFF
--- a/neurodsp/spectral.py
+++ b/neurodsp/spectral.py
@@ -507,37 +507,3 @@ def rotate_powerlaw(psd, f_axis, delta_f, f_rotation=30):
 
     # apply mask
     return psd * f_mask
-
-# def rotate_powerlaw(spectrum, f_axis, delta_f, f_rotation=30):
-#     """Takes a PSD and changes its power law exponent about a given axis (frequency).
-#
-#     Parameters
-#     ----------
-#     spectrum : array, 1-D
-#         Power spectrum to be rotated.
-#     f_axis : array, 1-D, Hz
-#         Frequency axis of input spectrum. Must be same dimension as spectrum.
-#     delta_f : float
-#         Change in power law exponent to be applied. Positive is counterclockwise
-#         rotation (flatten), negative is clockwise rotation (steepen).
-#     f_rotation : float, Hz
-#         Axis of rotation frequency, such that power at that frequency is unchanged
-#         by the rotation. Only matters if not further normalizing signal variance.
-#
-#     Returns
-#     -------
-#     x : array, 1-D
-#         Rotated spectrum.
-#
-#     """
-#
-#     # make the 1/f rotation mask
-#     f_mask = np.zeros_like(f_axis)
-#     f_mask[1:] = 10**(np.log10(np.abs(f_axis[1:])) * (delta_f))
-#     f_mask[0] = 1.
-#
-#     # normalize power at rotation frequency
-#     f_mask = f_mask / f_mask[np.where(f_axis >= f_rotation)[0][0]]
-#
-#     # apply mask
-#     return spectrum * f_mask
--- a/neurodsp/spectral.py
+++ b/neurodsp/spectral.py
@@ -502,10 +502,10 @@ def rotate_powerlaw(psd, f_axis, delta_f, f_rotation=None):
 
     if f_rotation is not None:
         # normalize power at rotation frequency
-        f_mask = f_mask / f_mask[np.where(f_axis >= f_rotation)[0][0]]
-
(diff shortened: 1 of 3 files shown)


neurodsp/spectral.py
```diff
@@ -439,36 +466,44 @@ def morlet_convolve(x, f0, Fs, w=7, s=.5, M=None, norm='sss'):
     return mwt_real + 1j * mwt_imag
 
 
-def rotate_powerlaw(spectrum, f_axis, delta_f, f_rotation=30):
-    """Takes a PSD and changes its power law exponent about a given axis (frequency).
+def rotate_powerlaw(psd, f_axis, delta_f, f_rotation=30):
```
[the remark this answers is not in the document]
rdgao: good point, I'll change default to 1Hz (which is the natural default of the transformation)

rdgao:
Yep. Well, I've written a test for it, and it passes. 

Sure, I can include a tutorial update

neurodsp/spectral.py
```diff
@@ -439,36 +466,46 @@ def morlet_convolve(x, f0, Fs, w=7, s=.5, M=None, norm='sss'):
     return mwt_real + 1j * mwt_imag
 
 
-def rotate_powerlaw(spectrum, f_axis, delta_f, f_rotation=30):
-    """Takes a PSD and changes its power law exponent about a given axis (frequency).
+def rotate_powerlaw(psd, f_axis, delta_f, f_rotation=None):
```
TomDonoghue: It seems weird to me to have input & output order as 'psd', 'freqs'. I'm used to the other order - but that could be just me. Is there a general order for NDSP? I did notice that this is inconsistent with the psd function, which is annoying. 

For example `rotate_powerlaw(*psd(signal, fs), 0.5, 30)` would not work, because the order will be wrong, and one might want this to work. It would also be nice to have things like this be consistent between modules. 

tl-dr: can we switch input & output order?

neurodsp/spectral.py
```diff
@@ -439,36 +466,46 @@ def morlet_convolve(x, f0, Fs, w=7, s=.5, M=None, norm='sss'):
...
+    f_rotation : float, Hz, optional
         Axis of rotation frequency, such that power at that frequency is unchanged
         by the rotation. Only matters if not further normalizing signal variance.
+        If None, the transform normalizes to power at 1Hz by defaults.
 
     Returns
     -------
-    x : array, 1-D
-        Rotated spectrum.
+    x : 1d array
```
TomDonoghue: Return is anonymous, so don't need the 'x' here.

neurodsp/spectral.py
```diff
@@ -439,36 +466,46 @@ def morlet_convolve(x, f0, Fs, w=7, s=.5, M=None, norm='sss'):
...
     """
-
     # make the 1/f rotation mask
     f_mask = np.zeros_like(f_axis)
-    f_mask[1:] = 10**(np.log10(np.abs(f_axis[1:])) * (delta_f))
-    f_mask[0] = 1.
+    if f_axis[0] == 0.:
+        # If starting freq is 0Hz, default power at 0Hz to old value because log
+        # will return inf. Use case occurs in simulating/manipulating time series.
+        f_mask[0] = 1.
```
TomDonoghue: Why does this do both of chasing the value at f[0], and ignoring the value at f[0] - doesn't it only really need to do one of these?

rdgao: Not sure if I'm understanding you here - it checks if f0 is 0Hz, and if so, manually set f_mask[0] to 1, and set the rest of f_mask to the computed value. Otherwise, f_mask[0] will return as infinity.

neurodsp/spectral.py
rdgao: 👍

neurodsp/spectral.py
rdgao: sure, good with me. Not sure if this comes up anywhere else, but we can make this consistent across.

neurodsp/spectral.py
TomDonoghue: Oops, sorry - misread it as if you were setting psd[0] = 1, but then skipping that value anyways. As you were :).

TomDonoghue: Yeh, looks good to me!

srcole: same

PR #80, opened 2018-10-28
Style & Doc Updates

TomDonoghue:
Some updates from a quick run-through with a linter, as well as some small doc updates. 

Updates:
- Cleans up imports: remove some unused, and re-order.
- Fix up some docstrings
- Linting fixes: line length, spacing, etc
- Small code refactorings, of things highlighted by linter

Notes:
- ToDo: fill in returns from `detect_bursts`
- In `detect_bursts` there is an input,  `filter_fn` that is never used. Is this a mistake?
- In burst, `_fit_slope` has an option to plot something - but matplotlib is never imported
	- We could of course import matplotlib here - but perhaps plots can/should be defined in a separate file?
- in swm/_find_new_windowix, there is an input `L_samp` that is never used

I also added a 'summary' shell script that I usually have in my projects, to check things like this. We don't have to merge that here necessarily - but I find it useful if you want to add it.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -107,3 +107,4 @@ venv.bak/
 .DS_Store
 neurodsp/*_sandbox.ipynb
 *_old
+_lint.txt
--- a/neurodsp/__init__.py
+++ b/neurodsp/__init__.py
@@ -4,6 +4,8 @@ from .filt import filter
 from .laggedcoherence import lagged_coherence
 from .timefrequency import phase_by_time, amp_by_time, freq_by_time
 from .burst import detect_bursts, detect_bursts_bosc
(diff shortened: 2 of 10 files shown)


TomDonoghue:
Question @srcole :
Why are a bunch of the parameters sim defaulted to `None`, and then set to specific values inside the function itself, instead of having the defaults directly defined?

Before getting distracted on more general stuff, I was initially just trying to figure out what the actual defaults are (specifically, what `prop_enter` and `prop_leave` end up if passed as `None` from `sim_noisy_bursty_oscillation`). It seems harder than it should be to see defaults in this case.

srcole:
Thanks for highlighting these issues and starting to fix them!

For the sim defaults, I forget exactly what I was trying to do, but I don't think it made a whole lot of sense. So yes, agree we should make it so those defaults are clear.

I'll start going through the PR and making some changes

srcole:
I think this is good to merge

rdgao:
agreed

Issue #81, opened 2018-10-28
Move plotting functions


TomDonoghue:
There is some plotting scattered around the module - sometimes plotting functions in other module scripts, or plots defined directly inside functions that other things.

This is perhaps not ideal:
- There is no easy way to see, overall what plotting functionality is available in the module
- Plots defined directly in functions break "1 function = 1 task", and also make them harder to apply (can't apply onto existing returned vars, without re-computing)
- Plot functions within specific sub-modules somewhat dilutes the focus of the module (it's not "task" but "task & related plotting").

Recommendation: can we move all plotting things to a `plts` module (or even, folder with multiple sections), and import as needed? Then everything still works as currently, but plots are more organized, and easier to find / apply / maintain. 

Note: If y'all are cool with this - I can do it.

rdgao:
good call - kinda ugly/scattered right now.

srcole:
yeah, i like the idea of a plotting module

srcole:
Will close this issue once plots are moved to their own separate folder, as discuss in #86

PR #86, opened 2018-10-30
Move plots to their own file

TomDonoghue:
WIP for #81 

What is here should be good. 

To do:
- Do we want to add tests for plots?
- Need to double check docs, maybe some updates there. 
- Do we wants to have a flat plts file, as is, or a folder of plot related things, so it would be (for example) `from ndsp.plts.spectral import plot_spectra`. This might have a cleaner organization in the long run.


srcole:
I think smoke tests for plots are good. But nothing more than that. Just show them in the tutorials.

rdgao:
separate plotting folder would be nice, actually. 

We don't have a lot of stuff right now, but I feel like thinking ahead would be beneficial so we don't have to go through this whole exercise again down the road.

TomDonoghue:
Yeh - agreed on having a folder, I'll update to that. 

Tests: yeh, I usually have tests that the plots at least run (don't really check what they plot, that's a bit hard). 

I'll update these.

TomDonoghue:
@srcole & @rdgao 

Can we merge this, and I'll make a separate PR on folder & tests?

Only reason - 'names' branched of here, so could be a conflict if they both go on separate paths when we try to bring them together again.

Issue #79, opened 2018-10-25
filter error unhelpful for transition bandwidth


rdgao:
currently, when a filter length is not long enough for computing transition bandwidth (I assume), it throws an unhelpful error, after the bandwidth warning. This occurs in bandstop filtering but presumably generalizes to the other options too for FIR.

From @tunmiseo 

=UserWarning: Error computing transition bandwidth of the filter. Defined filter length may be too short.
 raise warnings.warn(‘Error computing transition bandwidth of the filter. Defined filter length may be too short.’)
/code/lib/neurodsp/neurodsp/filt.py in filter(x, Fs, pass_type, fc, N_cycles, N_seconds, iir, butterworth_order, plot_frequency_response, return_kernel, verbose, compute_transition_band, remove_edge_artifacts)
   238                 warnings.warn(‘Transition bandwidth is ’ + str(np.round(transition_bw, 1)) + ' Hz. This is greater than the desired pass/stop bandwidth of ' + str(np.round(pass_bw, 1)) + ' Hz’)
   239         except StopIteration:
--> 240             raise warnings.warn(‘Error computing transition bandwidth of the filter. Defined filter length may be too short.’)
   241
   242     # Remove edge artifacts

*TypeError: exceptions must derive from BaseException*

PR #87, opened 2018-10-30
Fix transition band warnings

srcole:
Issue #79


srcole:
So the transition band computation was having problems cause I was too lazy to check if the frequency response ever indeed went below -20dB (the strong attenuation cutoff for the edge of the transition band).

I've updated the code so that it does this now and provides a warning. In the examples below, we see this warning appearing appropriately for a highpass and bandstop filter when the filters are defined to be too short.

image

image


image


image

srcole:
@rdgao @tunmiseo

rdgao:
looks good by eye, and it looks like that original unhelpful error is gone. So I'm cool with merge, and @tunmiseo can test it again.

Issue #83, opened 2018-10-29
clean bursting module


srcole:
Going through PR #80, I noticed a few changes I think should be made to prepare `burst.py` for v1.0
* `detect_bursts()`
    * Should be renamed to something like `detect_bursts_dual_amplitude_threshold` to be consistent with `detect_bursts_bosc`
    * Variable name improvements
        * X → signal
        * Min_osc_periods → min_cycles
        * Devation_type → average_method
    * Threshold parameters should be required
    * Remove algorithm input (only the median method)
    * Remove filter_fn optional input (unnecessary, and only complicates things)
* `detect_bursts_bosc()` I think this should either be labeled as experimental, private, or taken out since it really hasn't been tested or used much at all. Maybe just taken out cause it's a published method by another group (though there is no public code I can find for it)
* `get_stats()` - rename to something more descriptive like `compute_bursting_statistics` and could change a few things

srcole:
I think it would be good to leave BOSC so no one uses it thinking that it's a reliable implementation (because I don't remember testing it much). And if one of us really wants to implement BOSC, then they can just go back to the version of the code before we deleted it and start with it.

TomDonoghue:
I would say keep BOSC, if we have a working implementation of it (we can add more tests). Like, 'FOOOF is better than BOSC' shouldn't just be a sentence we say, it can also have 'and if you want to check, here is a script that compares them using the implementation in NDSP'. We can add a note in the docstring that BOSC is not the best option or something.

TomDonoghue:
Arbitrary filter support:
On this, I mostly defer to you both. I agree that I see no strong reason to support it. If it is already in there though, not clear we need to drop it. 

Also - our current filter function is a beast - it arguable does too many thing. What if we split our current filt func into sub-functions (still with a 'master' function that calls them), and then the pass in a filter function approach could also be a local function - makes more use of that approach.

PR #89, opened 2018-10-30
Clean bursting module

srcole:
Issue #83


srcole:
@TomDonoghue and @rdgao 
I wanted to make some changes as described in Issue #83. Mainly
* I think it's best to take out BOSC because it hasn't really been tested how well it works. And if someone wants to add a reliable BOSC function in the future, then they can go back to pre-v1.0 and build off of this code that we took out
* Changed some variable names and function names to be clearer

Any reasons not to go ahead and merge?

Issue #82, opened 2018-10-28
Variable names


TomDonoghue:
We don't have super consistent variable naming. With a breaking 1.0 release, should we go through and systematize it all to snake_case (so, for example, things like 'N_samples' -> 'n_samples'). 

Also, some specific cases:
- `filter` overwrites the python `filter` function, which can be really problematic.  
  - Especially because NDSP by default imports all functions directly in local namespace, this squashes filter always, meaning you can never use Python filter, and so has side effects / breaks things
- `psd`, as a function, strikes me as not a great name. Elsewhere, `psd` is used as a variable name, and it sounds like data more than a function
  - Heuristic: function names should be 'verb-y', because they do things.

We also have a bunch of single character variables, but this (in some cases), might be fine for the math-y parts, and also some names, like 'Fs', that are non-compliant, but (I think), are inherited from style in scipy.signal, or similar. Do we want to update those names too?

rdgao:
- could either change the name (use ```filter_ts()```) or don't do the auto imports in __init__, so it doesn't squash unless explicitly done.
- ```compute_spectrum``` ? 
- I'm good with all snake case, including changing Fs to fs.

rdgao:
do we have assignments for this? Are @srcole and I cleaning our respective modules, or somebody wants to just do it wholesale?

srcole:
if you can do `filt` and `spectral` and `sim`, I'll do `timefrequency` and `swm` and `laggedcoherence` and `burst`

TomDonoghue:
For consistency between files - wholesale might be better. 

I'm happy to do a first pass, on a branch, and y'all can scroll through and make any tweaks, if you want?

TomDonoghue:
This is a case where working collaboratively on a branch (as opposed to a fork) is probably useful.

TomDonoghue:
Like, this is a case where it might be good to coordinate. I made a branch with a first pass on 'filt'. Are we going with 'filter_signal', for the function name?

There are a couple more single char variables I would like to update (a, b, w, h) - but not sure best new names. Suggestions?

TomDonoghue:
Some particular suggestions - how do we feel about:
- x -> sig
- Fs -> freqs

(I'll update this as I scroll through.)

srcole:
a, b, w, h are common DSP standards, as used in `scipy.signal.freqz()`
https://docs.scipy.org/doc/scipy/reference/generated/scipy.signal.freqz.html

[...]

PR #90, opened 2018-10-30
Update naming

TomDonoghue:
Responds to #82 - update all the names.

Opened the PR to see if/what conflicts with other changes going in. 

Edit: should be pretty much there. I didn't bother with the tests, for now.


srcole:
Just scrolled through it. Good shit, Tom. Thanks so much for going through and improving all that naming!
Was there anything else you wanted to do?
I think it's good to merge.

rdgao:
yep, looks good to me. MVP effort.

TomDonoghue:
I did a last check on the tests last night too. From my perspective this should be good

Because so much code has changed, this is a PR that could use a bit of pushing on - if you want to clone this branch, and check that you can run things without anything changing, that would probably be good. I did also change more names than were explicitly mentioned in #82, so double check I didn't change anything that will annoy you :). 

After that, should be good to merge. This PR does fundamentally break consistency on master, so whenever people clone from now on, related code will need to be updated.

srcole:
The only thing I can think of is that we still have to change the orders of inputs in `sim.py` so that they're consistent across functions, as mentioned in issue #67. But I think that can be done in another PR to close this issue.

This is it for issue #82 , right?

And since all the tests pass that check for consistency with previous versions of the code, I think that tests pretty well that nothing has functionally changed. So I think it's good to merge. And then we can quickly tackle the rest of the issues before releasing 1.0

PR #94, opened 2018-10-31
Move plots to own folder

srcole:
Complete issue #81

does this match what you had in mind, @TomDonoghue ?


TomDonoghue:
Yep - this is what is what I was going for!

I might keep it as 'ndsp/plts/*' though, with the 's', partly because there are a lot of plots, and partly to clearly differentiate 'plts' from the common mpl abbreviation 'plt.

PR #96, opened 2018-11-01
Simulation code cleaning

srcole:
issue #67

including:
* adjust order of inputs to be consistent (`n_seconds`, `fs`, `freq`)
* allow noisy oscillator functions to have variable power law
* allow no filter for noise in noisy oscillator functions


srcole:
I think I addressed all the suggested changes to sim
Good to mere? @TomDonoghue @rdgao

neurodsp/sim.py
```diff
@@ -120,78 +99,89 @@ def sim_oscillator(n_samples_cycle, n_cycles, rdsym=.5):
...
+        if filter_order % 2 == 0:
+            filter_order += 1
 
     # Determine length of signal in samples
     n_samples = int(n_seconds * fs)
 
-    # Generate filtered brown noise
-    brown = sim_filtered_brown_noise(n_seconds, fs, (f_hipass_brown, None), filter_order)
+    # Generate filtered noise
+    noise = sim_filtered_noise(n_seconds, fs, f_range_filter, filter_order, exponent=exponent)
```
rdgao: do we want to pass in the different noise sim functions, instead of constraining to filtered noise? Because as of now this cannot produce the plateau/knee in time domain, as FOOOF is able to do in freq-domain.

rdgao:
changes look good to me. One note about using various noise generators for sim_noisy_oscillator though

srcole:
cool. I'll merge this then and open up a separate issue specifically for that

Issue #91, opened 2018-10-31
Test Coverage


TomDonoghue:
Missing test coverage:
- spectral.morlet_transform
- spectral.morlet_convolve
- burst.compute_burst_stats
- plts.*

srcole:
* sim.sim_poisson_pop
* sim.make_synaptic_kernel
* sim.make_osc_cycle

srcole:
I think the last ones needed are just the spectral ones. @rdgao

rdgao:
goddamn it.

PR #102, opened 2018-11-02
Plts and compute_burst_stats tests

srcole:
partial issue #91


srcole:
Not really sure why 3.4 is failing.

Can we just drop 3.4 from our tests and only support 3.5+

3.4 has started to give us a few misc problems.

TomDonoghue:
I dropped 3.4 on FOOOF, because I was using some things (simple string manipulation stuff) that wasn't supported in 3.4. Doesn't seem to have been a problem / no one has complained - so if you want to change the listed Python versions in setup.py, and stop it running on travis, I think that would be fine!

TomDonoghue:
Also - throw an eye on .travis in FOOOF to see how to add testing of 3.7 (for some dumb reason it's a bit different than normal, but we might as well test the newest version).

rdgao:
alright I'm gonna merge this and test 3.7 on a different PR

PR #107, opened 2018-11-04
swappable noise generator function in oscillator simulators

rdgao:
a few major changes:
- made private function (`_return_noise_sim()`)  to handle noise generation given a string function parameter and dict arg for noise params. 
- Function is called in `sim_noisy_oscillator()` and `sim_noisy_bursty_oscillator()`, which calls the above 
- also allow a custom noise time series which is just normalized and added to the oscillation simulation, after checking for same length 
- changed normalization from power (sum of squares) to variance due to offset in some noise generation functions
- updated docs and tests for above
- moved `filter_order` computation to within `sim_filtered_noise()`, defaults to None, which forces 3 times highpass filter length

#107


srcole:
what do you think about having a default for the noise function argument and kwargs?

rdgao:
I'm okay with that but I didn't know which to pick. I would lean towards synaptic?

rg.

srcole:
Yeah, that sounds good to me

PR #116, opened 2018-11-12
changed timefrequency functions to accept tuple with None for lowpass…

rdgao:
… or highpass as f_range

Because previously it's impossible to get amp_by_time, with the intention of just lowpassing, without an insanely long filter, since low-freq cutoff is required.


srcole:
looks good to me.
The only thing I would maybe add is a smoke test to make sure it runs when passing in `(None, fc)` or `(fc, None)`.

Issue #124, opened 2018-12-05
Error for some extreme simulation settings


srcole:
Right now, it's quite possible to break the simulation function `sim_bursty_oscillator` if you're pretty careless.

Particularly, in the below example I want an average rdsym of 0.3 and a standard deviation of rdsym of 0.1. Therefore, on each cycle, there's approximately a 1/1000 chance that an rdsym value will be simulated that's less than 0, which makes no sense.

A similar issue could arise for amplitude and period.

I think what we should do is re-generate random numbers for amp, period, or rdsym if a negative value is simulated. And then throw a warning to the user when this happens to tell them they should think about changing their parameters (because if this warning comes up a lot, then the actual distribution generated will be different from their intended distribution, or at least non-Gaussian).

```
# Simulate signal
n_seconds = 100
fs = 1000
freq_osc = 10
f_range = (6, 14)
f_range_noise = (1, None)
filter_order_noise = 3001
ratio_osc_var1 = .5
ratio_osc_var2 = 1
prob_leave_burst = .2
prob_enter_burst = .1
rdsym1 = .5
rdsym2 = .3
cycle_features = {'amp_std': .3,
                  'period_std': 10,
                  'rdsym_std': .1}

np.random.seed(1)
sig2, osc2, noise2 = sim_noisy_bursty_oscillator(n_seconds, fs, freq_osc,
                                  noise_generator='filtered_powerlaw',
                                  noise_args={'exponent': -2, 'f_range':f_range_noise,
                                              'filter_order':filter_order_noise},
                                  rdsym=rdsym2, cycle_features=cycle_features,
                                  ratio_osc_var=ratio_osc_var2,
                                  prob_enter_burst=prob_enter_burst, prob_leave_burst=prob_leave_burst,
                                  return_components=True)
```

PR #125, opened 2018-12-05
resimulate invalid cycle parameters

srcole:
issue #124

The script in this issue now runs and throws the following warning:

```
/gh/bv/neurodsp/neurodsp/sim.py:244: UserWarning: A cycle was simulated with invalid feature(s) for:
                        **amp ** (e.g. less than 0). A new cycle was
                        simulated to replace it. If this warning appears often, you should
                        change your simulation parameters.
  change your simulation parameters.'''.format(features_invalid))
/gh/bv/neurodsp/neurodsp/sim.py:244: UserWarning: A cycle was simulated with invalid feature(s) for:
                        **rdsym ** (e.g. less than 0). A new cycle was
                        simulated to replace it. If this warning appears often, you should
                        change your simulation parameters.
  change your simulation parameters.'''.format(features_invalid))
```

I think this is good. Because invalid parameters were only simulated twice during a 100 second simulation of 10Hz oscillation, resimulating a couple times won’t bias the distribution much at all.

### DIFF
--- a/neurodsp/sim.py
+++ b/neurodsp/sim.py
@@ -218,12 +218,31 @@ def sim_bursty_oscillator(n_seconds, fs, freq, rdsym=.5, prob_enter_burst=.2,
                     np.random.randn() * cycle_features_use['amp_burst_std']
                 current_burst_rdsym_mean = cycle_features_use['rdsym_mean'] + \
                     np.random.randn() * cycle_features_use['rdsym_burst_std']
-            period = current_burst_period_mean + \
-                np.random.randn() * cycle_features_use['period_std']
-            amp = current_burst_amp_mean + \
-                np.random.randn() * cycle_features_use['amp_std']
-            rdsym = current_burst_rdsym_mean + \
-                np.random.randn() * cycle_features_use['rdsym_std']
+
+            features_valid = False
+            while features_valid is False:
+                period = current_burst_period_mean + \
+                    np.random.randn() * cycle_features_use['period_std']
+                amp = current_burst_amp_mean + \
+                    np.random.randn() * cycle_features_use['amp_std']
+                rdsym = current_burst_rdsym_mean + \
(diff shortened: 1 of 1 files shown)


rdgao:
```
period = current_burst_period_mean + \
                np.random.randn() * cycle_features_use['period_std']
amp = current_burst_amp_mean + \
                np.random.randn() * cycle_features_use['amp_std']
rdsym = current_burst_rdsym_mean + \
                np.random.randn() * cycle_features_use['rdsym_std']
```

this is what it was before, and I realized we can't just replace with truncnorm at 0 because that will cut the standard normal draw in half.

What we actually want to do is probably just take use `max(0.,amp)`?

rdgao:
alright I'm actually not 100% sure what to do here: 

If we clip any of those 3 values at 0 after the appropriate scaling, and the cycle is set to be oscillating, does it screw up the data generation for that cycle later in the code. E.g., if period is 0? 

Do period, amp, and rdsym affect the subsequent cycle generation differently if they were 0?

If clipping at 0 is not okay, perhaps we should flip that cycle to be NOT oscillating if any of the params generate invalid values? Or have them take on another (default) value?

I still don't think having a while loop here is the best idea, as it could significantly increase run time. One option is to try 5 times and if the values are still invalid, error out and make the user change the arg values.

srcole:
I would opt for the last option you mentioned. If after 5 tries, we're still having problems, then the input parameters are funky.

Then we can return to them that we were commonly simulating negative values for feature _____. And tell them to perhaps modify the mean and st dev so that the defined distribution does not fall below 0 much

srcole:
Right, I could see maybe the user would want to modify it to 1 (or 0, depending on how it's coded), if they wanted to be very strict that the simulated distribution matches the defined one

rdgao:
sounds good

rdgao:
closing this, see PR #128

PR #126, opened 2019-01-17
demeaned simulated noise before adding to oscillation in sim_noisy_os…

rdgao:
…cillator and sim_noisy_bursty_oscillator, updated test data accordingly


TomDonoghue:
Post-hoc quick review (this is the one that was accidentally merged, right?) - this all looks fine to me!

rdgao:
yeah, I wrapped a bunch of different changes into it, but all related (not sure if that follows convention)

PR #128, opened 2019-01-21
resimulate invalid cycle parameters 

rdgao:
issue #124 

(closed #125 and made a fresh commit b/c cloning and pushing to that PR was a pain and the changes there are not accepted)


rdgao:
for record keeping at the same place:

for some extreme simulation settings for bursty_oscillator, randomly drawn values may fall outside of the acceptable range (e.g., amplitude < 0).

the solution we went with was to try to resimulate a number of times (function param, defaults to 5) and if the values are still invalid, error out and make the user change the parameter setting values.

srcole:
looks good to me

rdgao:
so good to merge?

Issue #127, opened 2019-01-19
Test data files


TomDonoghue:
I suggest we re-organize the test data files, consolidating into one procedure / script that can generate all the synthetic data for testing. This procedure can then be run at the beginning of running the test suite, and we don't actually have to package any data files with the module. 

Also: is there any benefit to the real data files that we run some tests on? Presumably the tests just test that it gets the same answer as last time (nothing changed), but we could presumably do this on synthetic data, right?

TomDonoghue:
Okay - we've basically moved on from the test files, to a new test procedure that generates data on the fly. Going in with #150 (and some details in #140 ) so I'm going to close this now.

PR #129, opened 2019-01-28
Update Tests - for 1.1

TomDonoghue:
This is a work in progress PR, partly related to #127 

Currently this just has a bunch of clean ups & orgs for testing. 

What I was originally looking into is potentially refactoring some of the tests, to be a little more dependent on checking properties of the signals, rather than shipping a bunch of data that is only used to check consistency. 

At the least, I think we can & should have a separate script that can be used to regenerate all the test data when we need it. 

Note: if there is more to do on this idea, we can add it in here, so this PR is not ready to merge yet.


rdgao:
agreed - and yeah I think we can replace all the real data with simulated data, if for nothing else other than not shipping the real data

rdgao:
do you want to task out some of this so I can help with it? on the project board 😮 ?

or we can meet at some point and figure out roughly what we want to do here

rdgao:
the projects tab beside the PR tab up top? It's like a Trello style thing, pretty sure you told me about this

TomDonoghue:
Okay - so I'm going call this one here on test updates that we can merge in for 1.1 - so this should be ready to merge, if it looks good. In the end it has relatively minor updates (doesn't include the test overhaul). 

I think the full update of the tests should wait for 2.0, because otherwise we are just going to have to reorg the tests then anyway, when some API things change.

rdgao:
yep, sounds good

Issue #99, opened 2018-11-01
Split general filter function into specialized functions


srcole:
We talked about how the current filter function is a bit of a mess to deal with because of the different requirements for bandpass, highpass, lowpass, bandstop.

After discussing, @TomDonoghue and @rdgao and I generally agreed that it would be cleaner if we split them up into different sub-functions that could alternatively be called in addition to the general function

srcole:
I started working on this.

If we split this up into sub-functions and want it to have the same input checking and functionality as the master function, we end up with A LOT of duplicate code, which I think is worse than how it is now.

The alternative is to make this into private functions that are called by the master function. However, there are many different parts of the master function that check what the filter type is (for raising errors, for designing the filter, for determining transition band). Therefore, approach either requires multiple private functions per filter type or still some duplicate code. In the former case, the general function is not much cleaner than it is now. In the latter case, there's again a large amount of duplicate code.

After looking into this, I think we should just keep the code as it is now and delete the issue.

Thoughts @TomDonoghue and @rdgao ?

TomDonoghue:
Merged in #132 . Closing.

PR #131, opened 2019-02-02
Updates for Sphinx

TomDonoghue:
Updates / fixes / clean ups related to the sphinx doc site. 

Includes:
- Adds the remaining tutorials in sphinx format (finishes #109) and cleans up the existing ones
    - Also removes the old notebooks, now that it's all ported
- Update to sphinx configs and layouts
- Updates to code documentation, to make it play nice with sphinx, and some small fixes / style things. 

Overall, this is a series of small fix ups to documentation and sphinx set ups, and adds the remaining tutorials. I did slightly switch the theme and API doc layout, if anyone has strong feelings on those. 

Ready for review - this again probably easier to check the doc site:
https://neurodsp-tools.github.io/neurodsp/

PR #132, opened 2019-02-03
Refactor Filter Code

TomDonoghue:
Responds to #99 

So I had a hack at the filter file / function, related to the discussion in #99. 

The sort of 'natural' organization that comes out of simply re-organizing the code as it was is to split up FIR & IIR filters, and also to split out some checking functions, as we as the compute transition band to their own function. As it is, this just moves around code, but doesn't copy any. 

With just a little more work, this will, I think, make the FIR & IIR functions directly usable, if one wants to skip right to them, and also the 'calc_transition_band' can definitely be refactored. There are a bunch of other small updates / changes that could be done too. But before I get too far down the rabbit hole, wanted to check in. 

What do y'all think of reorganizing in this way? cc: @srcole @rdgao 

Note: WIP - not ready to merge.


srcole:
I know I initially put in some support for IIR filters, but I honestly, don't think we should support them. Pretty sure Roemer would agree. He told me to always use FIR, and that's all I use. We should be forced to think about the filtering parameters we used each time we implement a filter, in order to recognize how it is impacting our signals.

But if we want to keep IIR support for backwards-compatibility and whatnot. Then, yeah, I skimmed the code, and it seems like a good reorganization. Much clearer. 👍

rdgao:
I like this very much more than the giant spaghetti filter function too

I have some vague sense that iir is good for bandstopping line noise, and that's one of the major use cases, though I don't remember where I learned that from, though I think it's a common enough thing that it's worth keeping, esp since all the components are already there? Not married to it either way though.

voytek:
If we already implement IIR, there's no reason to take it out. But if we strongly suggest only using IIR for, e.g. bandstop or line noise, then only make it an option for a bandstop function?

Force people into proper methods through gardenpathed code?

TomDonoghue:
I'm broadly agnostic to what we should have (I'm not the filter expert), but if there are real use cases, then yeh, we should keep it in (no reason to drop it). 

We should *suggest / recommend* what to do, but I'm not so keen on being super prescriptive (in terms of forcing the code to only do certain things). We already print a warning to recommend IIR on 'bandstop' only. 

One reason:
If IIR only runs for bandstop, then people can't even run code to explore it's properties in other cases. So that becomes us saying you have to listen to us, but on faith, without the opportunity to see why. 

I don't think this is the way to convince people. If it has problems, let's add a tutorial / example motivating why we suggest the things we do. 

More than a tool for analyses, I tend to think of NDSP as a tool for exploring the properties of methods, so on average I think my position is to include the things, and use NDSP to motivate if and how one should use them, but not force it - transparently open rather than forced into our current ideas.

voytek:
> If IIR only runs for bandstop, then people can't even run code to explore it's properties in other cases. So that becomes us saying you have to listen to us, but on faith, without the opportunity to see why.

Great point.

Anyway, if the code's there, I see no reason to _remove_ options from people.

TomDonoghue:
Okay - I think I've more or less finished the actual code refactor here. Most recent updates are a big overhaul on the transition band stuff. 

The tests all still pass and I've eyeballed that nothing changed in the tutorial either. But I know less than y'all about what these things are supposed to do - can you check through a bit, maybe run some examples, and make sure that it all seems to work? Also double check transition bands is doing what it's supposed to. 

I think the `fir_filt` and `iir_filt` should be callable directly - there might be a bit more work to make that the case. Any thoughts on variable ordering / function organization, etc, let me know and we can finish this up. Right now - docs aren't updated (some are changed in #131 , and I'll add docs for the new functions when things are more settled). 

Q: What's the significance / source of the numbers -20 & -3 for transition bands?

rdgao:
I think this needs @srcole's eyeballs. 
Also get_transition_bands docstring is currently "words words words"

neurodsp/filt.py
```diff
@@ -67,204 +66,216 @@ def filter_signal(sig, fs, pass_type, fc, n_cycles=3, n_seconds=None,
...
+        if iir:
+            return sig_filt, (b_vals, a_vals)
+        else:
+            return sig_filt, kernel
+    else:
+        return sig_filt
+
+
+def check_filter_definitions(pass_type, fc):
+    """Words, words, words."""
```
voytek: update docstring

neurodsp/filt.py
```diff
@@ -67,204 +66,216 @@ def filter_signal(sig, fs, pass_type, fc, n_cycles=3, n_seconds=None,
...
             raise ValueError('Two cutoff frequencies required for bandpass and bandstop filters')
 
         # Map fc to f_lo and f_hi
         f_lo, f_hi = fc
 
-    # for LP and HP, a tuple or int/float can be passed
-    #   if a tuple is passed, it's assumed (0,f_hi) for LP; (f_lo,Nyq) for HP
-    # i.e., highpass is a bandpass with second cutoff freq at Nyquist freq
+    # For LP and HP can be tuple or int/float
+    #   Tuple is assumed to be (0, f_hi) for LP; (f_lo, Nyq) for HP
```
voytek: `Nyq` should be `f_nyq`, right

neurodsp/filt.py
```diff
@@ -67,204 +66,216 @@ def filter_signal(sig, fs, pass_type, fc, n_cycles=3, n_seconds=None,
...
+
+    # Plot frequency response, if desired
+    if plot_freq_response:
+        plot_frequency_response(fs, b_vals, a_vals)
+
+    return sig_filt, b_vals, a_vals
+
+
+def fir_filt(sig, fs, pass_type, f_lo, f_hi, f_nyq, n_cycles, n_seconds, plot_freq_response):
+    """Words, words, words."""
```
voytek: update docstring

neurodsp/filt.py
```diff
@@ -67,204 +66,216 @@ def filter_signal(sig, fs, pass_type, fc, n_cycles=3, n_seconds=None,
...
-                """The designed filter (length: {:d}) is longer than the signal (length: {:d}).
-                The filter needs to be shortened by decreasing the n_cycles or n_seconds parameter.
-                However, this will decrease the frequency resolution of the filter.""".format(filt_len, len(sig)))
-
-    # Compute nyquist frequency
-    f_nyq = fs / 2.
+            filt_len = int(np.ceil(fs * n_cycles / f_lo))
+
+    # Force filter length to be odd
+    if filt_len % 2 == 0:
```
voytek: I'm not seeing a catch notifying the user that the filter order has been changed. If we don't have one, we should add one.

neurodsp/filt.py
```diff
@@ -67,204 +66,216 @@ def filter_signal(sig, fs, pass_type, fc, n_cycles=3, n_seconds=None,
...
-        # Compute the frequency response in terms of Hz and dB
-        if not iir:
-            b_vals = kernel
-            a_vals = 1
 
-        w_vals, h_vals = signal.freqz(b_vals, a_vals)
-        f_db = w_vals * fs / (2. * np.pi)
-        db = 20 * np.log10(abs(h_vals))
+def calc_transition_band(b_vals, a_vals, fs, f_lo, f_hi, f_nyq, pass_type):
+    """Words, words, words."""
```
voytek: update docstring

neurodsp/filt.py
```diff
@@ -67,204 +66,216 @@ def filter_signal(sig, fs, pass_type, fc, n_cycles=3, n_seconds=None,
...
+    f_db = w_vals * fs / (2. * np.pi)
+    db = 20 * np.log10(abs(h_vals))
 
-    # Remove edge artifacts
-    if not iir and remove_edge_artifacts:
-        n_rmv = int(np.ceil(filt_len / 2))
-        sig_filt[:n_rmv] = np.nan
-        sig_filt[-n_rmv:] = np.nan
+    # Confirm frequency response goes below -20dB (significant attenuation)
+    if np.min(db) >= -20:
```
voytek: -20 and -3 are _kind of_ magic numbers, and we might want to expose them to users to define, and set defaults in undefined.

neurodsp/filt.py
```diff
@@ -67,204 +66,216 @@ def filter_signal(sig, fs, pass_type, fc, n_cycles=3, n_seconds=None,
...
+
+        # Raise warning if transition bandwidth is too high
+        if transition_bw > pass_bw:
+            warnings.warn('Transition bandwidth is ' + str(np.round(transition_bw, 1)) + \
+                          ' Hz. This is greater than the desired pass/stop bandwidth of '\
+                           + str(np.round(pass_bw, 1)) + ' Hz')
+
+
+def get_transition_band(vec, sel, low, high):
+    """Words, words, words."""
```
voytek: update docstring

neurodsp/filt.py
srcole: fair. Yeah, I don't think there's any problem with making them extra kwargs.

neurodsp/filt.py
```diff
@@ -69,204 +68,215 @@ def filter_signal(sig, fs, pass_type, fc, n_cycles=3, n_seconds=None,
...
+                                              n_seconds, plot_freq_response)
+
+    # Compute transition bandwidth
+    if compute_transition_band and verbose:
+        if not iir:
+            a_vals, b_vals = 1, kernel
+        calc_transition_band(b_vals, a_vals, fs, f_lo, f_hi, f_nyq, pass_type)
+
+    # Remove edge artifacts
+    if not iir and remove_edge_artifacts:
```
srcole: This doesn't deal with the scenario where `iir = True` and `remove_edge_artifacts = True`. Of course, that's because we can't completely remove edge artifacts from IIR filters. But if the user passes in this input, we should throw a warning like "Edge artifacts are not removed for IIR filters"

srcole: Hah. Nvm. See that in line 152.

srcole:
Great stuff! After addressing Brad's stuff, it looks good to me.

neurodsp/filt.py
TomDonoghue: Yep, fixed.

TomDonoghue:
Yes, I wrote above that I hadn't updated the docstrings, as I was waiting to make sure sub-function organizing, parameter ordering etc were all settled.

neurodsp/filt.py
TomDonoghue: I'm not sure this is really needed... the user specifies the filter length in either number of cycles or number of seconds. We convert that to samples, and then then jitter this number by 1 value if it ends up being odd. 

For example, in an alpha filter (8-12 Hz), at default of 3 cycles, and sampling rate of 500 Hz, this trigger does change the filter length from 188 samples to 189 samples, which is the equivalent of changes the filter length from 376 ms to 378 ms. Or, generally, this changes the filter length by 1/fs, where the user has typically specified the filter length in n_cycles, and wouldn't know the exact length in seconds anyway. 

Is this degree of change something we really need to tell the user? A warning would trigger, on average, for every second filter call, and it seems like it might be more confusing/annoying than helpful. 

(I'm pretty sure this is how it all works - someone let me know if I'm missing something here. This changes the filter length, not the filter order).

neurodsp/filt.py
TomDonoghue: I added this as a set-able parameter in the sub-function that now checks filter properties, including transition bands. It's not currently added as an arg to all the top level filter functions, just because it's getting to be a lot of parameters. It could be added - or it could be a case of saying "call the transition function manually if you want to check custom transition bands"

TomDonoghue:
Another round - I finished the things that make the FIR & IIR functions directly callable - ended up including a bunch more refactoring, but I think it's all pretty nice now - much easier to call and use specific functions. 

This, in terms of code, should be about done, though default values perhaps need filling in for a bunch of parameters. Either here, or in #129 the tests should be updated to reflect the new organization. Testing this stuff should be easier with this organization. 

Note: from this refactor, we can fix up `plot_filter_response`, which now no longer needs to calculate the filter response itself. 

I replied to a couple comments above - otherwise, everyone cool with this organization?

Everything here I think is as updated as possible while not breaking the API. However, there are some tweaks that might be nice, that would break the API (simplest example - the verbose argument now doesn't even do anything, as everything that's printed is a warning, that are already squash-able).

There are a few others things across the package that could be tweaked - maybe we do an API breaking 2.0 release pretty soon?

Everyone cool with this as it is?

Note: Docstrings still aren't done - that'll go in last.

voytek:
Great work, @TomDonoghue!

neurodsp/filt.py
TomDonoghue: @voytek bump on this small note :)

TomDonoghue:
Docs filed in - this should actually be it now.

TomDonoghue:
It seems like we're all broadly on board here, so I'm going to merge this in. 

@srcole if you happen to have any more comments, you can still throw em here (or somewhere), and we can iterate before a tagged release!

PR #133, opened 2019-02-06
fixed typo in spectral.compute_spectrum

rdgao:
typo causes compute_spectrum to crash when an outlier_pct is specified


rdgao:
I'm just gonna merge this, no need for review for the one-liner, I don't think

Issue #140, opened 2019-02-25
Possible API Changes & Refactoring


TomDonoghue:
The following is a set of questions / suggestions for some API breaking updates, if we want to do that (presumably, in a 2.0 release or similar) and as well as some possible refactoring ideas. (This is just a dump of some things I noted down randomly, and am dumping here). 

Filter:
- [x] after the refactor, everything currently maintains the old API, but there are some names that are not great & parameter ordering could be cleaner.
- [x] We can update the plot_frequency_response to either change it's inputs, or at least directly use the filter function to compute_frequency_response (drop duplicate code). 

Sim:
- [x] Why do we use the word 'oscillator'? I find it a little confusing - oscillator implies a generator / generative process, but really were simulating oscillations / rhythms right? Can we change this?
- [x] Similarly, maybe we want to move away from the name 'noise'? I know some of the things we generate are 'statistical noise', but for consistency it might be nice to consolidate (across tools), on talking about periodic & aperiodic components.  
- [x] Sim is large and likely to grow. We could consider making this a folder with 'sim_periodic', 'sim_aperiodic' and 'sim_combined' or similar. 
- [x] Why does `sim_filtered_noise` make it's own filter, instead of using ndsp.filter?

Spectral:
- [x] The rotate function can be refactored, as in FOOOF, and maybe even moved to live with sim, which is where it's used. It has some funky param names that could / should change (delta_f -> delta_exponent, or similar).
- [x] There is some refactoring that would clean things up a little, for example, separating out a helper function for dealing with nperseg, that is currently a lot of duplicated code. 
- [x] There are different names and conventions used for outlier removal -> consolidate on same API

Time Frequency:
- [x] `_get_filter_passtype` feels like it could be moved to filter, as a helper function called `infer_passtype`?
- [x] remove_edge_artifacts could be split out as it's own function (copied code), or rather, use the same helper function that is defined in filter.

Utils [New]:
- [x] There are a bunch of helper / utility functions that overlap between modules (such as dealing with NaNs, dropping edges, etc), that could be split out into their own utilities file.

Tests:
- [x] Consolidate on a smoke test procedure, that generates data on the fly and tests for execution and as much sanity checking and accuracy testing as we can - drop test data files. 

[...]

PR #145, opened 2019-02-28
Updates to Spectral & TimeFrequency

TomDonoghue:
Non API breaking refactors and clean ups of the spectral & time frequency module. (Somewhat related to #140 - but nothing that breaks anything here). 

Main changes:
- Split up sub-function for settings checking, discarding outliers and sub-selecting from spectra
- Split out the different compute spectrum methods
- Get time frequency to borrow & use the utility functions that were split out in `filter` (get rid of duplicated code within `timefrequency`, and across the package. This includes moving & renaming a 'infer_passtype' from tf -> filt. 

Notably, this fixes a bug in which the old version would fail out if you tried to discard outliers in compute_spectrum with a 1d signal.

### DIFF
--- a/neurodsp/filt.py
+++ b/neurodsp/filt.py
@@ -571,7 +571,7 @@ def _remove_nans(sig):
     return sig_removed, sig_nans
 
 
-def _restore_nans(sig, sig_nans):
+def _restore_nans(sig, sig_nans, dtype=float):
     """Restore NaN values to the edges of a 1d array.
 
     Parameters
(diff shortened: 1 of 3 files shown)


neurodsp/filt.py
```diff
@@ -546,6 +546,37 @@ def compute_nyquist(fs):
 
     return fs / 2.
 
+
+def infer_passtype(fc):
```
rdgao: I poked around and it looks like only `timefrequency` uses infer_passtype. Should filter also use it, or is that unnecessary since the filter_type is specified?

TomDonoghue: As currently set up, filter doesn't really need to use it, I think, and yeh, only `timefrequency` actually uses it (which is where this code originally lived). 

I moved it to filter, partly because it seems like a filter related function - but also since in future other things *could* use it (for example, perhaps when passing in a range to filter aperiodic stuff in sim - so I thought it might get used more broadly when we do some 1.0 related updates.

rdgao: sounds good

TomDonoghue:
Alright - merging this one in!

Issue #143, opened 2019-02-27
Add MNE integration tutorial


voytek:
Per the code review during the _Journal of Open Source Software_ [review of our code](https://github.com/openjournals/joss-reviews/issues/1272), we need to add a tutorial showing how to integrate neurodsp with MNE.

TomDonoghue:
Done with #146

PR #146, opened 2019-03-02
Add example using with MNE

TomDonoghue:
This is a draft of an example for using NDSP together with MNE, related to a suggestion on the JOSS review, and as mentioned in #143 

The broad outline is straightforward (grab the MNE sample data, and start messing with it), but I realized it's not so obvious what to do / show in this example. 

So far this is a fairly trivial example of extracting a channel of interest, and checking for bursts. 

The main thing we would want to show, though, is using NDSP + MNE across multiple channels, and probably in an event related manner, and focusing on custom and interesting NDSP specialties (not showing things that MNE already does well, such as filtering). 

So, what quick & straightforward analyses with NDSP would be most useful / interesting to show with an MNE organization / dataset? Ideally a specific but cool analysis on epoched data, across multiple electrodes (channel clusters). 

The example data is an audio-visual whereby subjects detect a visual stimulus presentation:
https://martinos.org/mne/stable/manual/sample_dataset.html

### DIFF
--- a/doc/conf.py
+++ b/doc/conf.py
@@ -111,6 +111,7 @@ html_theme_options = {
     'navbar_links': [
         ("API", "api"),
         ("Tutorial", "auto_tutorials/index"),
+        ("Examples", "auto_examples/index"),
         ("GitHub", "https://github.com/neurodsp-tools/neurodsp", True)
     ],
 
@@ -239,9 +240,7 @@ epub_exclude_files = ['search.html']
 
 # -- Extension configuration -------------------------------------------------
 sphinx_gallery_conf = {
-     # path to your examples scripts
     'examples_dirs': ['../examples', '../tutorials'],
-     # path where to save gallery generated examples
     'gallery_dirs': ['auto_examples', 'auto_tutorials'],
     'within_subsection_order': FileNameSortKey,
     'backreferences_dir': 'generated',
--- a/examples/plot_mne_example.py
+++ b/examples/plot_mne_example.py
@@ -1,0 +1,93 @@
+"""
+Using NeuroDSP with MNE
+=======================
+
+This example explores some example analyses using NeuroDSP, integrated with MNE.
+"""
+
+###################################################################################################
+
+import numpy as np
+import matplotlib.pyplot as plt
+
+import mne
+from mne import io
+from mne.datasets import sample
+
+from neurodsp.spectral import compute_spectrum, trim_spectrum
+from neurodsp.burst import detect_bursts_dual_threshold
+
+###################################################################################################
+
+# Get the data path for the MNE example data
+raw_fname = sample.data_path() + '/MEG/sample/sample_audvis_filt-0-40_raw.fif'
+
+# Load the file of example MNE data
+raw = io.read_raw_fif(raw_fname, preload=True, verbose=False)
+
+###################################################################################################
+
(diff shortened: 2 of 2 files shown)


TomDonoghue:
@voytek I think this probably does need at least a little more, perhaps just showing the burst analysis across multiple channels. There are some questions above in the original post about what is most useful to show, because this is _very_ bare bones at the moment.

voytek:
Right, but I see all our examples and tutorials as works in progress. My review of the PR isn't "this is what I ultimately want to see" but rather "yes! this is where I want to see it go."

TomDonoghue:
So this is the main thing for the NDSP paper - and I do need some feedback / answers on how to finish it, from the questions in the original post.

srcole:
Agree burst detection is a good start. I would say shape analysis would be a good next step, but I guess we're not in bycycle ;)

* Sliding window matching is 1 possibility:
Use MNE to source-localize / ICA / whatever processing needed in order to get desired time series. Then identify the kernel with SWM. Though, this is probably just a worse version of [multivariate alphacsc](https://arxiv.org/abs/1805.09654v2)

* Simulate data and put it into an MNE object and then analyze it in MNE?

* Compute lagged coherence (or SCV) to get an index of rhythmicity at each node. Then plot that on a brain using MNE.

TomDonoghue:
@srcole 

Yeh - I agree an example with ByCycle & MNE would be good - but that would live over in ByCycle! (Once this version is done, we can use it as a template, and add something somewhat analogous to ByCycle). 

I don't want this example to get too complex (doing things like source projection), because that all gets a bit over the top - really the main purpose would be to show the core of using NDSP + MNE, and then you can always follow MNE examples to add things like source projection. Putting synthetic data into MNE objects feels a little weird. 

I do think the SCV or lagged coherence would be good. I hacked a bit at applying those measures to the sample data, but wasn't sure how to interpret the somewhat messy outputs. I think I'll go back to that - follow up on the bursting by running across multiple channels, and try and add a lagged coherence thing, and hopefully leave it at that for now.

rdgao:
I'd not touch SCV - no idea how to to interpret, at least officially. 
Lag coherence is good, we're also moving some stuff over from ByCycle right?
Spectral histogram is a pretty easy thing, and you get a viz for free, though that's single channel operation. 
Alternatively, you can take the example data and do rotate_powerlaw on it, and see how that changes the time series?

voytek:
Can we merge this? Since this is the last thing JOSS needs?

TomDonoghue:
It needs a little more to be a complete tutorial I think. 

I'll bump it up the ToDo list - from the discussion above a little bit more stuff with bursts and/or sliding windows should finish it out enough for a first version of this tutorial, to merge & finish up for JOSS.

TomDonoghue:
Okay, I extended this a bit, adding a lagged coherence analysis that shows some reasonable but I think sorta cool things, and plotting the results, and this I think makes it all good enough to merge as first version that demonstrates the main points. 

SideNote: this took a bit of a battle to some of the plotting and colorbars working relatively well together... if anyone knows the dark secrets of adding custom colorbars to MNE topo-plots, please share your magic spells, because I spent forever trying to tweak things (if you don't know how to fix it, you are also forbidden from making any comments about the placement or otherwise of the colorbars :))

PR #150, opened 2019-03-18
Updates for v2.0.0

TomDonoghue:
Responds to #140 

WARNING: This is an API breaking update. Don't merge until we're all sure we're on board, and also until after we've tagged and released a 1.1.0 release. 

Okay - this is getting fairly close (without the test overhaul yet). 

A couple open questions:
- `filt` uses the name `fc` for defining filter range, but other modules that call `filt`, for example `time-frequency` use the name `f_range`, which is a bit inconsistent. Should we consolidate on one name?
- What are the ideas for potentially updating aperiodic sim function names?

Note: not having updated the tests properly yet, I've just turned off a few tests where the old consistency tests became out of data, so confidence in this exact version should be relatively low. 

At this stage - perhaps not ready for a proper / formal review, but if people want to skim through - start checking if y'all are on board with the things that are changing, and also if there are things that seemed to be missed so far. Starting trying out using it / stress tests and ideas on testing overhaul also very welcome!


neurodsp/sim/aperiodic.py
```diff
@@ -191,40 +190,17 @@ def sim_filtered_noise(n_seconds, fs, exponent=-2., f_range=(0.5, None), filter_
...
-        # High pass filter
-        taps = signal.firwin(filter_order, f_range[0] / nyq, pass_zero=False)
-        noise = signal.filtfilt(taps, [1], noise)
-
-    # Band pass filtered
-    else:
-        taps = signal.firwin(filter_order, np.array(
-            f_range) / nyq, pass_zero=False)
-        noise = signal.filtfilt(taps, [1], noise)
+    noise = filter_signal(noise, fs, infer_passtype(f_range), f_range, **filter_kwargs)
```
rdgao: nice

TomDonoghue:
@rdgao - do you have thoughts defining filter ranges `fc` vs `f_range` (more context above), and/or thoughts suggestions on naming for aperiodic functions?

Also: when you plot a filter kernel, what do the axes represent? What should the labels be?

TomDonoghue:
Another question - why do we support passing in an array of values as an aperiodic signal into the combined functions? If you already have an aperiodic signal, in an array, then you can just simulate an oscillation and add it to the aperiodic (using our normalized by variance function, if desired). 

So - y'all cool with dropping array inputs to combined funcs?

rdgao:
looked through for the particular points you mentioned @TomDonoghue :

- I think morlet.py should be moved to timefrequency, cuz it returns a morlet spectrogram, which you can then average in time to create a power spectrum. 

- I don't think it's necessary to split spectrum.py and scv.py. They're both summary measures for a tf representation over time, and I don't imagine there will be a lot more to be added there. 

- Actually, spectral and scv are strictly downstream of tf functions, and we might want to be able to compute a psd using different methods of computing the spectrogram, i.e. plugging in different spectrogram methods into compute_psd. Same with the spectral histogram. I'm not sure what the best way to do this is while still keeping that the user can just pass in an univariate time series. Maybe something similar to how we do sim, passing in the tf method as an arg?

- in terms of namespace, I would just dump everything in the submodule namespace, at least the (common) user facing ones. I personally hate digging through layers of modules to grab a function, and it's unlikely that there will be a namespace collision. Any other reasons to separate them?

TomDonoghue:
Okay y'all - I this this is close now to the 2.0 version.

There are some refactors / updates to sim:
- Generally, I've refactored the periodic approach to consolidate on specifying and repeating kernels. - 
- This is nice & extendable, but does break a bit with how what is now called `sim_bursty_oscillation_features` simulates the varying features. If possible, it would be nice to refactor this function to use the different kernels too. 
- There are some quick notes / Qs in the code in `sim_powerlaw`. 
- Something is a little funky with `sim_jittered_oscillation`. tbh, I'm not totally clear on what this function is supposed to do. 

^ @rdgao : if you can have a check through these things, that would be really useful. 

For everything else, we've been broadly reviewing and talking through things as they've gone in. What I suggest then is if this looks about close enough, we merge and make a pre-release, and get people using this version for a bit, to fish out any bugs / problems with it before a proper major release.

rdgao:
> Okay y'all - this is almost it now for the 2.0 version. Or at least, with only a couple open questions w.r.t the newest updates, I don't otherwise really have anything left on my list of things to update.
> 
> There are some refactors / updates to sim:
> 
> * Generally, I've refactored the periodic approach to consolidate on specifying and repeating kernels. -
> * This is nice & extendable, but does break a bit with how what is now called `sim_bursty_oscillation_features` simulates the varying features. If possible, it would be nice to refactor this function to use the different kernels too.
> * There are some quick notes / Qs in the code in `sim_powerlaw`.
> * Something is a little funky with `sim_jittered_oscillation`. tbh, I'm not totally clear on what this function is supposed to do.
> 
> ^ @rdgao : if you can have a check through these things, that would be really useful.
> 
> For everything else, we've been broadly reviewing and talking through things as they've gone in. What I suggest then is if this looks about close enough, we merge and make a pre-release, and get people using this version for a bit, to fish out any bugs / problems with it before a proper major release.

okay I just saw there are 110 file changes so let's talk about this in lab tomorrow lol

TomDonoghue:
Okay @rdgao - I fixed up the main topics from the IRL code review, which is to fix how the aperiodic rotation deals with f=0, and temporarily remove periodic sims that don't work with oscillation kernels / the new approach - I'll open a branch to merge that back in with the new outline. 

Otherwise, everything in here has had an overview sign off, and this PR is getting ridiculous, so, now that 1.1.X is released, I'm going to merge this (officially breaking master from 1.0, and moving it to 2.0).

PR #167, opened 2019-09-05
[ENH] - Add DFA to a new aperiodic module

rdgao:
Add Detrended Fluctuation Analaysis. 

Edit from Tom:
- Eric & Ryan, I've tagged you into this PR, now that we're moving onto 2.2 & aperiodic additions. 
    - whenever y'all get a chance, Eric if you can check math & Ryan if you can check code, that would be great!


neurodsp/aperiodic/dfa.py
```diff
@@ -0,0 +1,145 @@
...
+        - computes the range of signal windows, divided by the standard deviation
+
+    Empirically, DFA seems to work for a larger range of exponent values.
+    """
+
+    # Get log10 equi-spaced scales and translate that into window lengths
+    t_scales = np.logspace(np.log10(min_scale), np.log10(max_scale), n_scales)
+    win_lens = np.round(t_scales * fs).astype('int')
+
+    # Calculate culmulative sum of the signal
```
TomDonoghue: I suggest that we move this line into each of the sub-functions, and pass in the original signal to them. It's a copied line, but it feels like it's really part of the computation (and I think it cleans up re-computing the signal in rescaled-range). 

Partly, this makes the the sub-functions more stand-alone, as you pass in an actual time series, and wouldn't need to pre-compute cumulative sum if trying to use directly.

neurodsp/aperiodic/dfa.py
```diff
@@ -0,0 +1,145 @@
...
+    deg : int, optional, default=1
+        Polynomial degree for detrending.
+
+    Returns
+    -------
+    det_fluc : float
+        Measured detrended fluctuaiton, as the average error fits of the window.
+    """
+
+    # Gather windows & vectorize, so we can call math functions in one go
```
TomDonoghue: This is actually the same procedure that is now a helper function in `neurodsp.utils.data`, called `split_signal`, so we can use that, which gets rid of some duplicated code.

neurodsp/aperiodic/dfa.py
```diff
@@ -0,0 +1,145 @@
...
+    -------
+    rs : float
+        Average rescaled range over windows.
+    """
+
+    # Gather windows & vectorize, so we can call math functions in one go
+    n_win = int(np.floor(len(sig) / win_len))
+    sig_rect = np.reshape(sig[:n_win * win_len], (n_win, win_len)).T
+
+    # get back the sig by taking the derivative of sig_walk
```
TomDonoghue: This is just a way to re-compute the original signal, right? Am I missing something more to it?

It seems like more work than is needed, as opposed to just passing in the original signal. I suggest we just pass in the original signal, compute cumulative sum in here (see comment above), and then we don't need to re-compute it. 

Also: in practice this re-computes the demeaned version of the original signal. Which doesn't matter for computing std, but just seems a little roundabout (harder to read / understand the code, I find). Again - mostly just checking I'm not missing something here.

TomDonoghue:
@rdgao : I refactored a bit to try and match NDSP format, and left a couple quick questions in a review. I want to sanity check I'm not missing anything before I go through with a couple other updates. Whenever you get a chance (no huge rush) throw an eye on the comments & sign off if things looks good. I'm happy to make the changes, if nothing looks weird. 

Only other open question is naming. 'fractal' or 'scale-free' feels a little broad, as many other measures relate to those properties. I don't want to call everything DFA, as it isn't all DFA. Is 'fluctuation analysis' an at all reasonable catch-all term for the measures here? Can we call this file 'fluctuations.py'?

neurodsp/aperiodic/dfa.py
rdgao: great!

neurodsp/aperiodic/dfa.py
TomDonoghue: Okay sounds good - moved the cumulative sum calculation into the sub functions.

neurodsp/aperiodic/dfa.py
TomDonoghue: Updated to use `split_signal`.

neurodsp/aperiodic/dfa.py
TomDonoghue: Okay - so I reworked it to pass in the original signal, demean and calculate cumsum, and then resplit the demeaned signal and calculate std from there - so it still uses demeaned signal to calculated std, but doesn't have to recompute the original signal from the cumsum anymore.

TomDonoghue:
Okay @rdgao - I did the updates from the comments here. One convo is left open, if you want to throw an eye and make sure I didn't mess it up in the updates. 

Other than potentially changing the name of the file, this should be ready to merge, I think, if everything looks good to you.

neurodsp/aperiodic/dfa.py
rdgao: yeah I think that looks right. I'm following the math from here, and it looks like things are done in the right order: https://en.wikipedia.org/wiki/Detrended_fluctuation_analysis

neurodsp/tests/test_aperiodic_dfa.py
```diff
@@ -0,0 +1,20 @@
...
+from neurodsp.tests.settings import FS
+
+from neurodsp.aperiodic.dfa import *
+
+###################################################################################################
+###################################################################################################
+
+def test_compute_fluctuations(tsig):
+
+    t_scales, flucs, exp = compute_fluctuations(tsig, FS)
```
rdgao: what's tsig?

rdgao:
Code looks okay. I think the test should run it on a simulated white noise signal and get back what we expect?

neurodsp/tests/test_aperiodic_dfa.py
TomDonoghue: It's just a pre-computed test signal. It's just a small optimization, from pytest, to not have to generate data objects every time you need one. 

It's generated in `conftest.py`, from:
```
@pytest.fixture(scope='session')
def tsig():

    yield np.random.randn(1000)
```

neurodsp/aperiodic/dfa.py
```diff
@@ -0,0 +1,143 @@
...
+    """
+
+    # Calculate cumulative sum of the signal & split the signal into segments
+    segments = split_signal(sp.cumsum(sig - np.mean(sig)), win_len).T
+
+    # Calculate local trend, as the line of best fit within the time window
+    _, fluc, _, _, _ = np.polyfit(np.arange(win_len), segments, deg=deg, full=True)
+
+    # Convert to root-mean squared error, from squared error
+    det_fluc = np.mean((fluc / win_len)**0.5)
```
elybrand, line 141: I believe this should be `np.mean((fluc / win_len))**0.5`.

neurodsp/aperiodic/dfa.py
```diff
@@ -0,0 +1,143 @@
...
+    # Get log10 equi-spaced scales and translate that into window lengths
+    t_scales = np.logspace(np.log10(min_scale), np.log10(max_scale), n_scales)
+    win_lens = np.round(t_scales * fs).astype('int')
+
+    # Step through each scale and measure fluctuations
+    fluctuations = np.zeros_like(t_scales)
+    for idx, win_len in enumerate(win_lens):
+
+        if method == 'dfa':
+            fluctuations[idx] = compute_detrended_fluctuation(sig, win_len=win_len, deg=deg)
```
ryanhammonds, line 73: This breaks when `win_len == 1`, causing `np.arange(win_len) ==  np.array([0]) ` to be the first arg in `np.polyfit`. Here is a minimal example:
```
import numpy as np
import scipy as sp
from neurodsp.utils.data import split_signal

tsig = np.random.randn(1000)

win_len = 1
segments = split_signal(np.cumsum(tsig - np.mean(tsig)), win_len).T
deg = 1

np.polyfit(np.arange(win_len), segments, deg=deg, full=True)
```

I'll have to dig into the code more to determine an appropriate fix for this.

elybrand: This seems like an edge case to me. DFA is just fitting a spline--that is, piecewise polynomials--to the data over intervals of length `win_len`. When `win_len ==1`, there's nothing to fit!

ryanhammonds: Thanks, we can add this to fix:
```
if method == 'dfa' and win_len > 1:
```
I'll push this tonight.

neurodsp/aperiodic/dfa.py
TomDonoghue: I'll take your word for it - updating!
In practice, it seems to make little difference.

neurodsp/aperiodic/dfa.py
TomDonoghue: A window length of 1 is a bad settings, given the requested scales and sampling rate. I don't think we should be implicitly skipping this, but rather throwing an error for bad inputs. 

It's only an issue in our tests because we use a low sampling rate (100), which is lower than any real signals, so in practice this shouldn't often be an issue. 

I'm going to update the tests, since for these measures we can check real values, and that should avoid the testing problem.

neurodsp/aperiodic/dfa.py
```diff
@@ -0,0 +1,143 @@
...
+        elif method == 'rs':
+            fluctuations[idx] = compute_rescaled_range(sig, win_len=win_len)
+        else:
+            raise ValueError('Fluctuation method not understood.')
+
+    # Calculate the relationship between between fluctuations & time scales
+    exp = np.polyfit(np.log10(t_scales), np.log10(fluctuations), deg=1)[0]
```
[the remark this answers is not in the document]
TomDonoghue: Yeh, I think this is a good point / idea - but is more of an extension for future work, that doesn't need to go into this PR necessarily. As I understand it, what we have here is "standard" DFA, which we can always build on later.

TomDonoghue:
I made some mild updates, based on the reviews - fixing the math error noted by Eric, updating tests to be more specific, as suggested by Richard, and updating the approach the window length issue brought up by Ryan (go team!). 

I think this should be good to go. @elybrand - maybe you want to take a quick look at my updates and how these functions are tested now? @ryanhammonds - everything new look good to you?

TomDonoghue:
Okay, all looks good here! Thanks for the code @rdgao, and the reviews @elybrand & @ryanhammonds!

Merging in now to officially start our aperiodic module!

rdgao:
wooo!

PR #182, opened 2020-05-17
Small module related updates

TomDonoghue:
A small set up of updates related to managing the module. 
Includes replaces the `summary.sh` file with a Makefile.

Note: NeuroDSP already had a Makefile, related to a Docker file that was contributed. 
I'm not that familiar with Docker, but just running it fails on the NeuroDSP install. The reason is right now it copies the `setup.py` and tries to run the copy, but that breaks reading in the version file. I'm not too clear on Docker, so I don't know much about why it does what it does, or how to fix it. So, it's an open question what to do with it. Maybe we just drop it? As far as I can tell, it's not really that standard for a python module repo to include a dockerfile (?).


Dockerfile
```diff
@@ -15,9 +15,4 @@ RUN conda create -n neurodsp python -q -y
 RUN ["/bin/bash", "-c", "source activate neurodsp && \
                         pip install -q jupyter && \
                         pip install -q -r requirements.txt && \
-                        python setup.py -q develop"]
-
-
-# If not in interactive mode, make the tutorials available over the exposed port
-COPY . /neurodsp
-CMD ["source activate neurodsp && jupyter notebook tutorials/ --ip=0.0.0.0 --port=5555 --no-browser --allow-root"]
+                        python setup.py -q develop"]
```
ryanhammonds, line 18: When building the image using the Dockerfile, I ran into this error:
```
Traceback (most recent call last):
  File "setup.py", line 7, in <module>
    with open(os.path.join('neurodsp', 'version.py')) as version_file:
FileNotFoundError: [Errno 2] No such file or directory: 'neurodsp/version.py'
```
It looks like docker doesn't like the relative paths that are referenced in the new setup.py. The solution I found was to make to copy the version.py and README.rst to the container, along with with it's structure (ie neurodsp/version.py and README.rst at the base level). This required renaming the conda env. Here is the diff that allowed me to successfully build:
```
diff --git a/Dockerfile b/Dockerfile
index 650430c..61a0fbd 100644
--- a/Dockerfile
+++ b/Dockerfile
@@ -8,11 +8,16 @@ EXPOSE 5555
 RUN apt-get update -qq && apt-get upgrade -qq -y && conda update conda -q -y
 WORKDIR /neurodsp
 
+# This lets setup.py find correct files:
+RUN mkdir -p /neurodsp/neurodsp
+COPY neurodsp/version.py /neurodsp/neurodsp/version.py
+COPY README.rst /neurodsp/README.rst
+
 # Create the conda environment and install dependencies
 COPY requirements.txt /neurodsp/requirements.txt
 COPY setup.py /neurodsp/setup.py
-RUN conda create -n neurodsp python -q -y
-RUN ["/bin/bash", "-c", "source activate neurodsp && \
+RUN conda create -n env python -q -y
+RUN ["/bin/bash", "-c", "source activate env && \
                         pip install -q jupyter && \
                         pip install -q -r requirements.txt && \
                         python setup.py -q develop"]
```

TomDonoghue:
Hey @ryanhammonds : thanks for digging into the Dockerfile, and figuring out a fix - your edits all look good and make sense to me. 

I guess I'm still left with questions on the purpose of this file though:
- Is it usual to have to copy files around in Docker files to get them to install?
     - This setup still feels sorta weird to me.
- What is the use case for a Dockerfile that builds the development version of NDSP?
    - Is this useful to other people? Is it a relatively general use case?

I'm a little unclear on these questions, so I'm wondering if there it is worth keeping the Docker file, or perhaps just dropping it. I can't think of or find examples of other modules maintaining Dockerfiles, for development versions, and I my guess is that anyone who knows Docker, and wanst an image with NDSP, can more easily write a Dockerfile that install the stable version. Am I missing something?

TomDonoghue:
@ryanhammonds - I think maybe this fell through the cracks a bit, so whenever you get a chance, do you have any thoughts on the Qs about the Dockerfile above? TBH, since I don't think we ever use the Docker, and it's not clear to me it's all that useful to others, my inclination is just to drop it. 

If you have any comments, let me know, and I'll make a decision / edit from there, and then we can merge this.

Dockerfile
TomDonoghue: Thanks for digging in - as discussed, Dockerfile is now dropped!

doc/glossary.rst
```diff
@@ -17,13 +18,16 @@ General
 Digital Signal Processing
 -------------------------
 
-For a general introduction to digital signal processing, we recommend
-`Seeing Circles Sines and Signal <https://jackschaedler.github.io/circles-sines-signals/>`_
-by Jack Schaedler.
+Digital Signal Processing (DSP) is a vast area of science & engineering of computational analysis
+to digital signals.
```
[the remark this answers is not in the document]
TomDonoghue: Yeh, good catch - updated!

ryanhammonds:
Sorry I'm just now responding! Feel free to tag me in for a re-review if you'd like.

PR #185, opened 2020-06-02
[FIX] Allow line styles when plotting.

ryanhammonds:
Before,  'alpha' and 'lw' had no effect when plotting with `plot_time_series`, and this PR fixes this. A list can be passed for either line style argument, and the style will be applied to each line in the corresponding order. For example, in `plot_time_series(times, [sig_a, sig_b], lw=[1, 5])`, a linewidth of 1 will be applied to sig_a and a width of 5 for sig_b. Alternatively a single value can be passed and all lines will have the same style.

Here is a [notebook](https://github.com/ryanhammonds/neurodsp/blob/explores/line_styles.ipynb) that has an example of the change.

### DIFF
--- a/neurodsp/plts/settings.py
+++ b/neurodsp/plts/settings.py
@@ -11,7 +11,7 @@ LINE_STYLE_ARGS = ['alpha', 'lw', 'linewidth', 'ls', 'linestyle']
 # Custom style arguments are those that are custom-handled by the plot style function
 CUSTOM_STYLE_ARGS = ['title_fontsize', 'label_size', 'tick_labelsize',
                      'legend_size', 'legend_loc']
-STYLERS = ['plot_styler', 'line_styler', 'custom_styler']
+STYLERS = ['axis_styler', 'line_styler', 'custom_styler']
 STYLE_ARGS = AXIS_STYLE_ARGS + LINE_STYLE_ARGS + CUSTOM_STYLE_ARGS + STYLERS
 
 ## Define default values for aesthetic
--- a/neurodsp/plts/settings.py
+++ b/neurodsp/plts/settings.py
@@ -1,0 +1,22 @@
+"""Default settings for plots."""
+
+###################################################################################################
+###################################################################################################
+
+## Define collections of style arguments
+# Plot style arguments are those that can be defined on an axis object
+PLOT_STYLE_ARGS = ['title', 'xlabel', 'ylabel', 'xlim', 'ylim']
+# Line style arguments are those that can be defined on a line object
+LINE_STYLE_ARGS = ['alpha', 'lw', 'linewidth', 'ls', 'linestyle']
+# Custom style arguments are those that are customly handled by the plot style function
+CUSTOM_STYLE_ARGS = ['title_fontsize', 'label_size', 'tick_labelsize'
+                      'legend_size', 'legend_loc']
+STYLE_ARGS = PLOT_STYLE_ARGS + LINE_STYLE_ARGS + CUSTOM_STYLE_ARGS
+
+## Define default values for aesthetic
+# These are all custom style arguments
+TITLE_FONTSIZE = 20
+LABEL_SIZE = 16
+TICK_LABELSIZE = 16
+LEGEND_SIZE = 12
+LEGEND_LOC = 'best'
--- a/neurodsp/plts/settings.py
+++ b/neurodsp/plts/settings.py
@@ -5,13 +5,14 @@
 
 ## Define collections of style arguments
 # Plot style arguments are those that can be defined on an axis object
-PLOT_STYLE_ARGS = ['title', 'xlabel', 'ylabel', 'xlim', 'ylim']
+AXIS_STYLE_ARGS = ['title', 'xlabel', 'ylabel', 'xlim', 'ylim']
 # Line style arguments are those that can be defined on a line object
 LINE_STYLE_ARGS = ['alpha', 'lw', 'linewidth', 'ls', 'linestyle']
 # Custom style arguments are those that are custom-handled by the plot style function
 CUSTOM_STYLE_ARGS = ['title_fontsize', 'label_size', 'tick_labelsize',
                      'legend_size', 'legend_loc']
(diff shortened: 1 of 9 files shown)


TomDonoghue:
Okay @ryanhammonds - thanks for noticing and proposing a fix for this issue. 

I started a review, and found myself writing point by point comments for a bunch of things to suggest a re-org, and I realized it would probably save us both some time if I just pushed the suggestions. So I've updated some things. The main goal, is to keep (and now document) the distinction between `plot_style` and `styple_plot` (perhaps, in hindsight, not particularly well named). `style_plot` is a decorator, and shouldn't do any styling itself - it organizes inputs and passes things through, applying a function to tune styling after plot creating. By default, this is `plot_style`. So I moved all the style application to `plot_style` (and refactored it a bit). You can see this most clearly in the first commit (subsequent commits or just a bit of extra stuff). 

The goal / point is that if one wants to use our plots, but define custom styling, you can write and pass in a custom styler function, for example (and this should work - try it in your notebook):
```
def custom_styler(ax, **kwargs):
    ax.set_title('Some squiggles')
    ax.set_xlabel('My special units', {'fontweight' : 'bold'})
plot_time_series(times, [sig_filt, amp], colors=['k', 'r'], plot_style=custom_styler)
```

None of this was actually really documented. Ooops.  I at least tried to document the functions a bit better. As it stands, still there isn't really any way for a casual user to know this, so it should probably be mentioned the docs somewhere. 

The extra commits are just a bit of extra stuff, that gives us access to passing in custom style arguments. For example, this should also work in your notebook example:
```
plot_time_series(times, [sig_filt, amp, sig_filt+0.25, amp+0.25], colors=['k', 'r'],
                 title='Double Trouble', ylabel='Oscillation amplitude', lw=[5, 1],
                 alpha=[0.5, 0.5, 0.25, 0.25], title_fontsize=20, tick_labelsize=20)
```

The good news is that nothing I changed impacts the API, or how things work from your original commit - so this doesn't change anything built on top of plotting (this just extends it a bit). We can chat on how any of this works, if you want. If you want to give this a check over, and review, I think it should be about ready to merge.

ryanhammonds:
Hi @TomDonoghue, I'm digging into your changes now to figure out how everything is working. Your description makes sense, but the behavior I'm seeing in a notebook seems unintended. Should a `custom_styler` overwrite all default plot style settings?

This changes x/y axis limits and the dimensions of the figure, without those setting specified:
```
def custom_styler(ax, **kwargs):
    ax.set_title('Some squiggles')
    ax.set_xlabel('My special units', {'fontweight' : 'bold'})

plot_time_series(times, sig, colors=['k'], plot_style=custom_styler)
```
compared to default behavior:
```
plot_time_series(times, sig, colors=['k'])
```

TomDonoghue:
Hey @ryanhammonds - so yeh, as it was, replacing `plot_style` completely overwrote any styling, and so any `kwargs` passed in didn't get applied. That was how it was set up. 

You comment got me thinking more that thats not really ideal. So, probably totally over-doing it at this point, a new commit splits out different plot style applications. One can still wholesale replace `plot_style`, as before (the example above should still work the same). But the newest commit also allows for this:
```
def custom_styler(ax, **kwargs):
    ax.set_title('Some squiggles')
    ax.set_xlabel('My special units', {'fontweight' : 'bold'})
plot_time_series(times, [sig_filt, amp], colors=['k', 'r'], alpha=[0.5, 0.25],
                 custom_styler=custom_styler)
```
^In this case, the `custom_styler` is specifically replaced, allowing for customization, but still calling the (now split-up) default functions for applying axis & line styling. 

Does this make sense, do you think?

ryanhammonds:
I'm confused by the behavior of the latest commit and want to make sure my confusion isn't a bug.

Should custom_styler leave unspecified styling as default? When I tested, custom_styler plots the same thing as plot_style.

Using line_styler does what I would expect axis_styler (and maybe custom_styler) to do.

```
def custom_styler(ax, **kwargs):
    ax.set_title('Some squiggles')
    ax.set_xlabel('My special units', {'fontweight' : 'bold'})

# These will plot the same thing
plot_time_series(times, sig, colors=['k'], custom_styler=custom_styler)
plot_time_series(times, sig, colors=['k'], plot_style=custom_styler)

# This leaves default axis styling alone and only changes the specificed axis changes.
plot_time_series(times, sig, colors=['k'], line_styler=custom_styler)

# This raises an error. I expected this to produce what is in the line above
plot_time_series(times, sig, colors=['k'], axis_styler=custom_styler)
```

TomDonoghue:
hey @ryanhammonds - you are right, that last example should have run, and didn't because of a name error - now fixed. 

The `axis_styler` and `line_styler` are technically available if a user wants to overwrite the default `apply_axis_style` or `apply_line_style`. This is likely not really needed. It just sort of comes for free when splitting out - I think passing a default `custom_styler` could be useful, while still be able to also apply any general axis or line stylings. Point being, if it was unclear, those last two plot examples in the comment above, though they both now run, do not really reflect expected usage. 

An example of the difference between using plot_style and custom_styler:
```
# These will plot the same thing
#   This is since `plot_style` overwrites all style (voiding the alpha input)
#    By comparison `custom_style` allows adding customization, still calling apply for axis & line
plot_time_series(times, sig, colors=['k'], alpha = 0.5, custom_styler=custom_styler)
plot_time_series(times, sig, colors=['k'], plot_style=custom_styler)
```

ryanhammonds:
Hi @TomDonoghue, coverage of the files in this PR are now at 100%.

There was a bug that I fixed in the latest commit too: `plot_power_spectra` uses `itertools.cycle` when defining colors, but the function wasn't imported.

TomDonoghue:
Hey @ryanhammonds - I've updated the tests for the plots. 

I realized I wasn't very clear on 'adding tests' and while you added did increase the coverage, I think it's better to try and maintain more modularity / specificity for tests (one test per function), rather that adding things to be called indirectly. In this way, if something breaks, it should hopefully be easier to track down where this problem arises from. Since I didn't describe this very well, and it's tricky to ask to sketch out tests for weird code that someone else wrote, I thought it would be easier for me to just show this and organize tests for the style code. 

If you want to have a look through my last commit - if everything looks good to you, then I think this should be ready to merge!

Issue #174, opened 2020-01-16
Discontinuities when simulating periodic signals in neurodsp.sim


nschawor:
Stationary simulated periodic signals display some discontinuities.
<img width="676" alt="Screen Shot 2019-12-17 at 10 23 04" src="https://user-images.githubusercontent.com/1282430/72542834-a02fcd00-3839-11ea-92a7-e01c79331775.png">

Using the first three blocks of example code from here: https://neurodsp-tools.github.io/neurodsp/auto_tutorials/sim/plot_SimulatePeriodic.html

TomDonoghue:
Okay, so this has been bugging me, and I wasn't sure if there was anything we could do to fix it, within the framework of simulating periodic components as concatenations of cycles (as we do).

This mismatch arises from uneven sampling, as interactions between the requested sampling rate and frequency of the cycle. If the signal sample rate doesn't play nice with the frequency, the cycles don't concatenate well. 

So, my idea / suggestion is to sample cycles at the nice sampling rate for them, build the signal, and then resample the whole signal to the desired sampling rate at the end. 

So, something like:
```
def sim_oscillation_test(n_seconds, fs, freq, cycle='sine', **cycle_params):

    # Figure out how many cycles are needed for the signal, & length of each cycle
    n_cycles = int(np.ceil(n_seconds * freq))
    # UPDATE: Change how `n_seconds_cycle` is computed (old version was weird)
    n_seconds_cycle = 1/freq 

    # UPDATE: Tune sampling rate for simulating cycle
    # Create oscillation by tiling a single cycle of the desired oscillation
    #   The cycle is created a sampling rate that is a multiple of the frequency
    #   This is done to create cycles that line up well for concatenation
    cycle_fs = freq * 100
    osc_cycle = sim_cycle(n_seconds_cycle, cycle_fs, cycle, **cycle_params)
    
    # Create full signal by tiling the simulated single cycle
    sig = np.tile(osc_cycle, n_cycles)
    
    # UPDATE: resample to request sampling rate
    # Resample cycle to desired sampling rate
    sig = resample(sig, int(n_cycles * 1/freq * fs))

    # Truncate the length of the signal to be the number of expected samples
    #   This is done because we simulate an integer number of cycles, 
    #   which can be beyond the requested data, given the freq & n_seconds input
    n_samps = int(n_seconds * fs)
    sig = sig[:n_samps]
    
    return sig
```

#### First pass tests
[...]

PR #189, opened 2020-06-15
OLD / DON'T MERGE - Fix concatenation issues in periodic simulations

TomDonoghue:
Responds to #174 

Updates:
- change to how cycles are sim'd & concatenated, in order to tune the sampling rate to the cycle to concatenate without 'edge effects' between concatenated cycles, that cause issues. This is now a new 'tilable cycle' function that gets used to sim oscillations & bursts. 
- updated the normalization decorator, to more specifically only apply the 'select-nonzero' to to burst signals. This allows for scaling signals, but in the case of bursts, keeping non-bursts timepoints at zero

For review:
@rdgao : if you don't mind, since you know this code and we've talked about this before, I'd appreciate if you could throw an eye on the updates and spotcheck for anything you think looks potentially problematic and/or theoretically dubious
@ryanhammonds : please check through as much as you can that the code makes sense, but also for this one, I think a bunch of 'stress testing' is a good idea. Simulate lots of signals, using different types of cycles, and check through them for any issues. Make sure we don't have the discontinuities that were the original issue, and that we haven't introduced some new oddity. You can calculate power spectra on sim'd signals to check that frequencies line up, and that power spectra look as expected, etc. 

EDIT: nvm. I think this is the wrong solution (thanks to notes from Ryan!). Pause on this PR, I'm going re-evaluate and likely try again on this.

### DIFF
--- a/neurodsp/sim/aperiodic.py
+++ b/neurodsp/sim/aperiodic.py
@@ -15,7 +15,7 @@ from neurodsp.sim.transients import sim_synaptic_kernel
 ###################################################################################################
 ###################################################################################################
 
-@normalize
+@normalize()
 def sim_poisson_pop(n_seconds, fs, n_neurons=1000, firing_rate=2):
     """Simulate a Poisson population.
 
@@ -66,7 +66,7 @@ def sim_poisson_pop(n_seconds, fs, n_neurons=1000, firing_rate=2):
     return sig
 
 
-@normalize
+@normalize()
 def sim_synaptic_current(n_seconds, fs, n_neurons=1000, firing_rate=2.,
                          tau_r=0., tau_d=0.01, t_ker=None):
     """Simulate a signal as a synaptic current, which has 1/f characteristics with a knee.
@@ -117,7 +117,7 @@ def sim_synaptic_current(n_seconds, fs, n_neurons=1000, firing_rate=2.,
     return sig
 
 
-@normalize
+@normalize()
 def sim_random_walk(n_seconds, fs, theta=1., mu=0., sigma=5.):
     """Simulate a mean-reverting random walk, as an Ornstein-Uhlenbeck process.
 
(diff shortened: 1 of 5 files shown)


TomDonoghue:
@ryanhammonds - The main issue isn't the normalization, it's the discontinuities noted in the original issue. It does seem that the resampling (or something else) doesn't play nice with some of the cycle types.... I'll have to look into that.

ryanhammonds:
I thought it was normalization since the discontinuities are fixed in master if the normalize decorator is commented from `sim_oscillation`. I'll read the issue thread more carefully.

TomDonoghue:
> I thought it was normalization since the discontinuities are fixed in master if the normalize decorator is commented from `sim_oscillation`. I'll read the issue thread more carefully.

THEY ARE!? 

Well... damn. 

I think I've been chasing the whole bug this whole time, misunderstanding what's happening. I thought it was something to do with the concatenation, because there were off by one errors or something. 

So.... the whole resampling thing might just be irrelevant then. Whoooooops. This is actually sorta good news - as the resampling doesn't generalize to different cycle types. My last hacky commit actually drops the resampling anyways. 

I think I need to come back and check this again, and probably restart a PR, because now this one seems probably wrong-headed. 

Good catch Ryan! Thanks for figuring it out - I'm grateful for the fresh eyes!

rdgao:
so...everything is solved here?

also, from Ryan's plot, it seems like the new method introduces a ringing edge artifact, presumably from the antialiasing filter resample automatically applies. 

Have you guys observed that this is a consistent thing? If so, that might be an issue

TomDonoghue:
@rdgao - the problem is figured out, but this PR is the old / bad version! I'll tag you in to sanity a new & fixed version, but otherwise you can basically ignore this PR. 

Sidenote: the ringing in Ryan's graph is because the standard scipy Fourier based resampling doesn't play nice with non-sinuosoidal cycles. Luckily, the actual fix seems like it doesn't need any resampling, so I think we can sidestep that whole thing.

rdgao:
okay cool, sounds good, lmk

PR #190, opened 2020-06-19
[ENH] Color cycling defaults

ryanhammonds:
This changes the color default in `plot_time_series` to 'k' and 'r' for 1 or 2 signals. If more than two signals are given, then mpl's default color cycle is used. Users can still overwrite these defaults.

### DIFF
--- a/neurodsp/plts/time_series.py
+++ b/neurodsp/plts/time_series.py
@@ -60,8 +60,8 @@ def plot_time_series(times, sigs, labels=None, colors=None, ax=None, **kwargs):
         colors = ['k', 'r']
     colors = repeat(colors) if not isinstance(colors, list) else cycle(colors)
 
-    for time, sig, label in zip(times, sigs, labels):
-        ax.plot(time, sig, color=next(colors), label=label)
+    for time, sig, color, label in zip(times, sigs, colors, labels):
+        ax.plot(time, sig, color=color, label=label)
 
     ax.set_xlabel('Time (s)')
     ax.set_ylabel('Voltage (uV)')
--- a/neurodsp/plts/time_series.py
+++ b/neurodsp/plts/time_series.py
@@ -55,17 +55,14 @@ def plot_time_series(times, sigs, labels=None, colors=None, ax=None, **kwargs):
     else:
         labels = repeat(labels)
 
-    if colors is not None:
-        colors = repeat(colors) if not isinstance(colors, list) else cycle(colors)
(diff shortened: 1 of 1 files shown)


neurodsp/plts/time_series.py
```diff
@@ -57,11 +57,15 @@ def plot_time_series(times, sigs, labels=None, colors=None, ax=None, **kwargs):
...
-    else:
-        colors = cycle(['k', 'r', 'b', 'g', 'm', 'c'])
+    elif len(sigs) <= 2:
+        colors = cycle(['k', 'r'])
+
+    for time, sig, label in zip(times, sigs, labels):
 
-    for time, sig, color, label in zip(times, sigs, colors, labels):
-        ax.plot(time, sig, color, label=label)
+        if colors is not None:
```
TomDonoghue, line 65: Ahh, I see, because we use zip, we can't just leave the singular None. 

It's a little pedantic, but I would suggest updating above to:
```
# If not provided, default colours for up to two signal to black & red
if not colors and len(sigs) <= 2:
    colors = cycle(['k', 'r'])
colors = repeat(colors) if not isinstance(colors, list) else cycle(colors)
```

This way, a colors of None gets the repeat call, and we need less code overall to do this (we don't have to update the loop). Does that make sense?

ryanhammonds: The issue with this is that ax.plot doesn't like None. This will raise an error:
```ax.plot(times, sig, None)```

TomDonoghue: Oh, it does need the explicit keyword. It does some kind of "what is this" guess procedure for what `arg` inputs are (because it supports variable APIs), but `color=None` works, I think.

So, the plot call should be valid as:
`ax.plot(time, sig, color=color, label=label)`

ryanhammonds: Using None as a kwarg works, updated!

TomDonoghue: awesome!

So I direct committed a small tweak we don't need to double on calling cycle, or use the else, right?

ryanhammonds: Yeah, that's better. It functions the same and saves a line.

TomDonoghue: I went for one more small direct edit - editing back to the original & consistent zip approach.

TomDonoghue: Tests pass! This looks good to me.

PR #197, opened 2020-06-23
Extend to new line styling, and update application

TomDonoghue:
This PR extends application of line styling to allow it to include `marker` and `markersize` (1st commit). 

Relatedly, `plot_style` approach is updated to only apply line style to lines added to the plot in the current plot call (3rd commit).

The reason for the update to line style application is that the old approach would apply single style args to all lines even if they came from different plot calls. For example, in the old approach, in the following call both lines would be made dashed, which I think is unexpected / not ideal:
```
fig, ax = plt.subplots(figsize=(15, 3))
plot_time_series(times, sig1, lw=6, ax=ax)
plot_time_series(times, sig2, linestyle='--', ax=ax)
```

With the update, a line styling will only get applied to the line or lines from the plot call from which the styling was triggered, which I think makes more sense.

### DIFF
--- a/neurodsp/plts/settings.py
+++ b/neurodsp/plts/settings.py
@@ -7,7 +7,8 @@
 # Plot style arguments are those that can be defined on an axis object
 AXIS_STYLE_ARGS = ['title', 'xlabel', 'ylabel', 'xlim', 'ylim']
 # Line style arguments are those that can be defined on a line object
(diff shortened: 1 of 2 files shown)


TomDonoghue:
Sounds good! I also sanity checked that nothing goes weird on the docsite with this, so I think we should be good. I'll merge.

PR #200, opened 2020-07-02
[BUG] - Technical Audit: Spectral

elybrand:
This PR addresses two minor bugs in the spectral submodule. Specifically,

1. The wavelet spectrum now returns the average modulus squared of the wavelet coefficients, rather than the average wavelet coefficients. In other words, it returns an average power spectrum which is consistent with the return types of the other spectrums.
2. `compute_wavelet_spectrum` also now averages over the correct axis, reflecting the changes in the `compute_wavelet_transform` where frequency is now plotted along the y-axis. I had to incorporate this change which is a duplicate from the issue193_tf PR because I missed this downstream change.
3. Since all power spectral 2D arrays are now consistently formatted with frequency along the y-axis, I removed a few calls which transposed these arrays which previously had accounted for the formatting issue.

### DIFF
--- a/neurodsp/spectral/power.py
+++ b/neurodsp/spectral/power.py
@@ -106,7 +106,7 @@ def compute_spectrum_wavelet(sig, fs, freqs, avg_type='mean', **kwargs):
         freqs = create_freqs(*freqs)
 
     mwt = compute_wavelet_transform(sig, fs, freqs, **kwargs)
-    spectrum = get_avg_func(avg_type)(mwt, axis=0)
+    spectrum = get_avg_func(avg_type)(abs(mwt)**2, axis=1)
 
     return freqs, spectrum
 
--- a/neurodsp/timefrequency/hilbert.py
+++ b/neurodsp/timefrequency/hilbert.py
@@ -26,11 +26,11 @@ def robust_hilbert(sig, increase_n=False):
     Returns
     -------
     sig_hilb : 1d array
-        The Hilbert transform of the input signal.
+        The analytic signal, of which the imaginary part is the Hilbert transform of the input signal.
 
     Examples
     --------
-    Compute a Hilbert transform of a signal, using zero padding:
+    Compute the analytic signal, using zero padding:
 
     >>> from neurodsp.sim import sim_combined
     >>> sig = sim_combined(n_seconds=10, fs=500,
@@ -194,13 +194,23 @@ def freq_by_time(sig, fs, f_range=None, hilbert_increase_n=False,
     >>> instant_freq = freq_by_time(sig, fs=500, f_range=(8, 12))
     """
 
-    pha = phase_by_time(sig, fs, f_range, hilbert_increase_n,
-                        remove_edges, **filter_kwargs)
+    # Calculate instantaneous phase, from which we can compute instantaneous frequency
+    pha = phase_by_time(sig, fs, f_range, hilbert_increase_n, remove_edges, **filter_kwargs)
 
-    phadiff = np.diff(pha)
-    phadiff[phadiff < 0] = phadiff[phadiff < 0] + 2 * np.pi
+    # If filter edges were removed, temporarily drop nans for subsequent operations
+    pha, nans = remove_nans(pha)
 
-    i_f = fs * phadiff / (2 * np.pi)
+    # Differentiate the unwrapped phase, to estimate frequency as the rate of change of phase
+    pha_unwrapped = np.unwrap(pha)
+    pha_diff = np.diff(pha_unwrapped)
+
+    # Convert differentiated phase to frequency
+    i_f = fs * pha_diff / (2 * np.pi)
+
+    # Prepend nan value to re-align with original signal (necessary due to differentiation)
     i_f = np.insert(i_f, 0, np.nan)
+
+    # Restore nans, to re-align signal if filter edges were removed
+    i_f = restore_nans(i_f, nans)
 
     return i_f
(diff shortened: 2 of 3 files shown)


ryanhammonds:
I found the bug and the site now build correctly. Feel free to rebase and drop the commit if it has any consequences.

neurodsp/spectral/power.py
```diff
@@ -106,7 +106,7 @@ def compute_spectrum_wavelet(sig, fs, freqs, avg_type='mean', **kwargs):
         freqs = create_freqs(*freqs)
 
     mwt = compute_wavelet_transform(sig, fs, freqs, **kwargs)
-    spectrum = get_avg_func(avg_type)(mwt, axis=0)
+    spectrum = get_avg_func(avg_type)(abs(mwt)**2, axis=1)
```
TomDonoghue, line 109: This could perhaps use some in-code commentary. I'm thinking something like:

```
# Convert the wavelet coefficient outputs to power
mwt_power = abs(mwt)**2
# Create the power spectrum by averaging across the time dimension
spectrum = get_avg_func(avg_type)(abs(mwt)**2, axis=1)
```

My comments are totally ad-hoc / placeholders, in so far as I understand what's being done. If you think the description could be useful, please replace with your own that I'm sure will be more descriptive / useful!

neurodsp/spectral/variance.py
```diff
@@ -148,7 +148,7 @@ def compute_scv_rs(sig, fs, window='hann', nperseg=None, noverlap=0,
 
         # Repeated sub-sampling of spectrogram randomly, with replacement between draws
         for draw in range(ndraws):
-            idx = np.random.choice(spg.shape[1], size=nslices, replace=False)
+            idx = np.random.choice(spg.shape[1], size=nslices, replace=True)
```
TomDonoghue, line 151: Hmm, good catch, this is totally a disconnect between comment & actual code. 

I'm tagging in @rdgao to confirm that the doc is correct, and this sampling is intended to be replacement, and so code should be updated here.

neurodsp/spectral/variance.py
```diff
@@ -233,17 +233,17 @@ def compute_spectral_hist(sig, fs, window='hann', nperseg=None, noverlap=None,
     freqs, _, spg = spectrogram(sig, fs, window, nperseg, noverlap, return_onesided=True)
 
     # Get log10 power & limit to frequency range of interest before binning
-    ps = np.transpose(np.log10(spg))
+    ps = np.log10(spg)
```
TomDonoghue, line 236: So, these changes are related to the update in `trim_spectrum`, which I suggested we revert. 

I think something relevant is that `trim_spectrum` is conceived to be for a collection of spectra (organized as [n_spectra, n_freqs]), where as here we using spectrogram, and so have windows over time (organized as [n_freqs, n_times]]). 

These are two different cases that I think we want to maintain. 

One of the quirks here then, seems to me to arise from somewhat 'hacking' `trim_spectrum` to be used for a different purpose, leading to the transposing. I wonder if one way to clean this up is to add a `trim_spectrogram` utility function, which assumes the [n_freqs, n_times] orientation, and allows one to sub-select in either frequency or time. 

If so, we can add these utility, revisiting this function and updating what gets called here internally to make it a little more consistent.

TomDonoghue:
Okay, I've done my review here! Comments attached to code. @elybrand - please check through, and let me know what you think. 

I think a general idea here is that we want to have clean, and separate handling of handling spectra vs. handling spectrograms. I want to check what I wrote in code generally makes sense, and if so, we can update to change this PR a little accordingly.

neurodsp/spectral/variance.py
elybrand: I realized shortly after this commit that this doesn't affect anything. This is drawing one sample from the list of indices, so whether or not use draw with replacement the distribution is unaffected. Apologies for the confusion.

neurodsp/spectral/utils.py
```diff
@@ -46,8 +46,9 @@ def trim_spectrum(freqs, power_spectra, f_range):
 
     # Restrict freqs & psd to requested range. The if/else is to cover both 1d or 2d arrays
     freqs_ext = freqs[f_mask]
+
     power_spectra_ext = power_spectra[f_mask] if power_spectra.ndim == 1 \
-        else power_spectra[:, f_mask]
+        else power_spectra[f_mask, :]
```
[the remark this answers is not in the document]
elybrand: Thank you for the detailed comment. Basically the reason I changed some of the slicing and transposing is because `scipy.signal.spectrogram` outputs a 2D array where the horizontal slices are signals across time, and the vertical slices are spectra. I made this commit shortly after noticing `compute_wavelet_transform` had this *flipped*, so I thought it was an error.

Here is an example that I was imagining would be the typical use case.

```python
from neurodsp.spectral import trim_spectrum
from scipy.signal import spectrogram
import numpy as np


n_seconds = 5
fs = 100

# Create a sinusoid of frequency 1.
sig = sim_oscillation(n_seconds, fs, 1, n_cycles=5)

freqs, t_bins, spg = spectrogram(sig, fs, nperseg=100)
fmask = range(10)

freqs_ext, spg_ext = trim_spectrum(freqs, spg, [0,10])

# Only look at the first 10 frequencies.
plt.pcolormesh(t_bins, freqs_ext, spg_ext, shading='gouraud')
```

neurodsp/spectral/variance.py
elybrand: This makes perfect sense and points out my earlier confusion. Adding a `trim_spectrogram` seems like the cleanest way to resolve this.

neurodsp/spectral/power.py
elybrand: I think this is a great suggestion.

Issue #165, opened 2019-09-03
Wavelet normalization


TomDonoghue:
For wavelets, right now we have two normalizations available:
```
* 'sss' - divide by the square root of the sum of squares
* 'amp' - divide by the sum of amplitudes
```

And these are implemented, as written in the code. The question is that, somewhere through the history, the thing fell out of sync, and the 'amp' used to say, in the docs, that this was all divided by two, even though the code did not do this. 

What is somewhat unclear (to me, at least) is what the best wavelet normalizations are. Maybe it should be divided by two? So if anyone knows the secrets of wavelets, please throw out some info :).

TomDonoghue:
@elybrand - since you've just been looking at `timefrequency`, I thought I would tag you in here, as you might know the answer, and if there is any problem here, or if we can choose. 

The current wavelets have a couple possible normalizations. Basically, at some point we ended up with a discrepancy between the docs and the code. Without trying to dig out all the history of the code, I wondered if you had any quick thoughts on this, and on whether there is any reason normalizing by the amplitudes should be divided by 2 (which we don't currently do, but the docs used to say we did).

elybrand:
@TomDonoghue 

So I have a couple of remarks. First, I just realized there is no way the user can specify the normalization method when they call compute_wavelet_transform. It just defaults to 'sss'!

Second, I don't immediately see why you would need to divide by 2. My understanding of the two normalizations is as follows:

**L2 normalization**
If you were to select a wavelet transform where the wavelets you used formed a basis, normalizing by the L2 norm ('sss') would preserve the energy of the signal like the normalized DFT. Otherwise, continuous wavelet transforms, normalized or not with 'sss', do not preserve energy (numerically). Graphically **this does have an affect on the TF plot**. Wavelets at scale s have an L2 norm of 1/sqrt(s). Consequentially, if your signal were sig = sin(5x) + sin(10x), though both frequencies have equal amplitude, the magnitude of the wavelet transform at frequency 5 will be larger than that at 10 because the normalization factor. While it appears that the scaling, and therefore the norm, is fixed for the implementation of compute_wavelet_transform as is, it's actually not in practice. This is obfuscated by the fact that the wavelet length is dependent on which frequency convolve_wavelet is fed in. The morlet wavelet constructed in convolve_wavelet isn't actually the length of the signal. I imagine you did this to truncate the Gaussian envelope to zero.

**L1 normalization**
[...]

PR #201, opened 2020-07-07
Update Wavelets

elybrand:
This PR changes the default normalization of `compute_wavelet_transform` to `'amp'`. It also adds a note to the docstring which specifies that it is a continuous wavelet transform.


TomDonoghue:
Just adding a tracking note that this update stems from discussion in #165 

This looks good to me - merging in!

Issue #193, opened 2020-06-22
Tech-Audit #1: Code & Analysis


TomDonoghue:
This issue is part of a 'technical audit' of `neurodsp`, with @elybrand. 

The first task / goal is to check through the code and in particular, all the analysis functions. For the relevant sub-modules, we primarily want to check that everything, as implemented, is technically sound, and that the settings and implementations are appropriate for our purposes. It's also an opportunity to identify any places we could add extensions to functionality, and/or update and replace current approaches, if there appear to be better options available. 

In the case of `filt`, `timefrequency`, and `spectral`, these sub-modules implement basic and standard DSP procedures for the application to neuroscientific signals. The goal here is to check that these functions perform as expected, and look through to check for any possible issues with the current implementations - doing a check & audit for correctness. 

In the case of `burst` and `rhythm`, these sub-modules implement particular algorithms that have been proposed for the analysis of neuroscientific data. In this case, I think it makes sense to perhaps have a look at these functions and, in so far as you can, check that the implementations seem technically valid (that there are no mistakes in the calculations). This doesn't need to address whether these algorithms are ideal / well designed for there stated tasks (that's somewhat out of scope for this review). 

How you go about doing this is largely up to you! As you mentioned, I think using some simulations and checking that outputs are all as expected makes sense (these can be collected and later used to augment our code tests). For anything that comes up, feel free to comment it here, start a chat on Slack, and/or open a new issue for anything we should address.

elybrand:
So I finished reading through the filter tutorial and submodule. I have a few comments, which I'll list in decreasing urgency.

Your method of implementing an IIR filter risks running into some serious numerical instability. In iir.py, apply_iir_filter(), you call the filtfilt method. For high-order butterworth filters, this could lead to disaster. Instead, you should use sosfiltfilt(). This requires only minor modifications. Namely, when calling scipy.signal.butter(), use output='sos'. Feed this, instead of the b values and a values, into sosfiltfilt.

Here's a minimal working example, inspired from the filter tutorial, which shows how these two methods diverge when the butterworth order is 5.

```python
import numpy as np
from neurodsp.filt import filter_signal
from neurodsp.sim import sim_combined
from neurodsp.utils import create_times
from neurodsp.plts.time_series import plot_time_series
from scipy.signal import butter, sosfiltfilt

# Set the random seed, for consistency simulating data
np.random.seed(0)

# General setting for simulations
fs = 1000
n_seconds = 5

[...]

PR #202, opened 2020-07-07
[BUG] - Technical Audit: Spectral

elybrand:
After comments from the previous PR, this updated PR is pretty short. It basically fixes the incorrect calculation of how power is computed for the wavelet spectrum.

### DIFF
--- a/neurodsp/spectral/power.py
+++ b/neurodsp/spectral/power.py
@@ -214,7 +214,7 @@ def compute_spectrum_medfilt(sig, fs, filt_len=1., f_range=None):
     freqs = np.fft.fftfreq(len(sig), 1. / fs)[:int(np.ceil(len(sig) / 2.))]
 
     # Convert median filter length from Hz to samples, and make sure it is odd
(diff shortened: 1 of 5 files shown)


neurodsp/spectral/power.py
```diff
@@ -106,7 +106,10 @@ def compute_spectrum_wavelet(sig, fs, freqs, avg_type='mean', **kwargs):
         freqs = create_freqs(*freqs)
 
     mwt = compute_wavelet_transform(sig, fs, freqs, **kwargs)
-    spectrum = get_avg_func(avg_type)(abs(mwt)**2, axis=1)
+    # Convert the wavelet coefficient outputs to power
+    mwt_power = abs(mwt)**2
```
ryanhammonds, line 110: Thanks for the comment! It made things clearer.

TomDonoghue:
Okay, this looks good to me!

Note for tracking - this PR responds in general to #193, and was originally opened in #200, with some comments there that directed the updates here.

PR #203, opened 2020-07-07
[ENH] - Spectral: Trim Spectrogram

elybrand:
This PR adds `trim_spectrogram` which allows a user to trim particular frequency and time values from a spectrogram.


TomDonoghue:
This looks great! Thanks for the addition @elybrand! Merging in.

Issue #194, opened 2020-06-22
Tech-Audit #2: Code Tests


TomDonoghue:
This issue is part of a 'technical audit' of `neurodsp`, with @elybrand. 

Our current code tests are, broadly speaking, 'smoke tests', meaning they execute code in the module, and generally check that nothing breaks - but they generally do not test very much for accuracy of the outputs.

As part of the technical audit, it would be nice to update / extend the tests we have to do more accuracy checking. This might help to discover if there are any issues with our current implementations, as well as offering a better test suite to guide future development. 

If you are using simulations to check expected output in #193, this this issue can broadly be addressed by collecting these examples and trying to add them to our code tests, in so far as we can do so effectively and efficiently. 

All our tests currently work on simulated data, for which approximate results can be predicted. Testing analysis functions is quite difficult - so I think the goal here is to add more accuracy checking, if & where we can, but without "over-doing it" (as in, what can we add to be more confident in our implementations, and to be safer w.r.t. future development, without falling too deep into the rabbit hole of broadly difficult questions about how to precisely test code that computes estimations).

TomDonoghue:
@elybrand updated some of our tests to be more stringest in https://github.com/neurodsp-tools/neurodsp/pull/204

Overall, I think in terms of the audit this was checked, and there are no current ToDos on this. Closing.

PR #204, opened 2020-07-07
[ENH] - Tech Audit #2: Accuracy checks for Hilbert Transform, Filters, and Spectra

elybrand:
This PR adds accuracy checks (#194)  for

1. All hilbert.py related functions using a simple sinusoid.
2. FIR filtering of a low-frequency sinusoid. Specifically, it checks that it is attenuated under a high pass filter, and is minimally attenuated under a lowpass filter.
3. `compute_spectrum` with the `medfilt` and `welch` options. It uses a low frequency sinusoid to compare against where the power spectrum should be a dirac spike at frequency 1 and zero elsewhere.

### DIFF
--- a/neurodsp/tests/conftest.py
+++ b/neurodsp/tests/conftest.py
@@ -6,6 +6,8 @@ import numpy as np
 
 from neurodsp.utils.sim import set_random_seed
 
+from neurodsp.sim import sim_oscillation
+
 ###################################################################################################
 ###################################################################################################
 
@@ -22,3 +24,10 @@ def tsig():
 def tsig2d():
 
     yield np.random.randn(2, 1000)
+
+@pytest.fixture(scope='session')
+def sig_sine():
+
+	n_seconds = 1
+	fs = 100
+	yield (n_seconds, fs, sim_oscillation(n_seconds, fs, 1, variance=None, mean=None))
--- a/neurodsp/tests/conftest.py
+++ b/neurodsp/tests/conftest.py
@@ -5,6 +5,7 @@ import pytest
 import numpy as np
 
 from neurodsp.utils.sim import set_random_seed
+from neurodsp.tests.settings import FS, N_SECONDS, FREQ_SINE
 
 from neurodsp.sim import sim_oscillation
 
@@ -26,8 +27,6 @@ def tsig2d():
     yield np.random.randn(2, 1000)
 
 @pytest.fixture(scope='session')
-def sig_sine():
+def tsig_sine():
 
-	n_seconds = 1
-	fs = 100
-	yield (n_seconds, fs, sim_oscillation(n_seconds, fs, 1, variance=None, mean=None))
+	yield sim_oscillation(N_SECONDS, FS, freq=FREQ_SINE, variance=None, mean=None)
--- a/neurodsp/tests/conftest.py
+++ b/neurodsp/tests/conftest.py
@@ -5,7 +5,7 @@ import pytest
 import numpy as np
 
 from neurodsp.utils.sim import set_random_seed
-from neurodsp.tests.settings import FS, N_SECONDS, FREQ_SINE
+from neurodsp.tests.settings import FS, N_SECONDS, N_SECONDS_LONG, FREQ_SINE
 
 from neurodsp.sim import sim_oscillation
 
@@ -30,3 +30,8 @@ def tsig2d():
 def tsig_sine():
 
 	yield sim_oscillation(N_SECONDS, FS, freq=FREQ_SINE, variance=None, mean=None)
+
+@pytest.fixture(scope='session')
+def tsig_sine_long():
+
+	yield sim_oscillation(N_SECONDS_LONG, FS, freq=FREQ_SINE, variance=None, mean=None)
--- a/neurodsp/tests/test_filt_filter.py
+++ b/neurodsp/tests/test_filt_filter.py
@@ -2,7 +2,7 @@
 
 from pytest import raises, warns
 
-from neurodsp.tests.settings import FS, FILT_EPS
+from neurodsp.tests.settings import FS, EPS_FILT
 
 from neurodsp.filt.filter import *
 from neurodsp.filt.filter import _iir_checks
@@ -12,7 +12,7 @@ import numpy as np
 ###################################################################################################
 ###################################################################################################
 
-def test_filter_signal(tsig, sig_sine):
+def test_filter_signal(tsig, tsig_sine):
 
     out = filter_signal(tsig, FS, 'bandpass', (8, 12), filter_type='fir')
     assert out.shape == tsig.shape
@@ -24,16 +24,15 @@ def test_filter_signal(tsig, sig_sine):
         out = filter_signal(tsig, FS, 'bandpass', (8, 12), filter_type='bad')
 
     # Apply lowpass to low-frequency sine. There should be little attenuation.
-    n_seconds, fs, sig = sig_sine
-    sig_filt_lp = filter_signal(sig, fs, pass_type='lowpass', f_range=(None, 10))
+    sig_filt_lp = filter_signal(tsig_sine, FS, pass_type='lowpass', f_range=(None, 10))
     # Compare the two signals only at those times where the filtered signal is not nan.
     not_nan = ~np.isnan(sig_filt_lp)
-    assert np.max(np.abs(sig[not_nan] - sig_filt_lp[not_nan])) < FILT_EPS
+    assert np.max(np.abs(tsig_sine[not_nan] - sig_filt_lp[not_nan])) < EPS_FILT
 
     # Now apply a high pass filter. The signal should be significantly attenuated.
-    sig_filt_hp = filter_signal(sig, fs, pass_type='highpass', f_range=(30, None))
+    sig_filt_hp = filter_signal(tsig_sine, FS, pass_type='highpass', f_range=(30, None))
     not_nan = ~np.isnan(sig_filt_hp)
-    assert np.max(np.abs(sig_filt_hp[not_nan])) < FILT_EPS
+    assert np.max(np.abs(sig_filt_hp[not_nan])) < EPS_FILT
 
 
 def test_iir_checks():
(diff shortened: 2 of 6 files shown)


TomDonoghue:
Okay, slightly unconventional review approach here. I thought I would do some updates, so I started doing directly, then realized I had a bit more to suggest - so I did some direct edits, then I'm also leaving some review comments. 

Please take a look at my direct commit, to sanity check, and get a sense of what I was going for. Basically, that commit is trying to consolidate on using general test settings and refactoring some down to minimize a bit.

neurodsp/tests/conftest.py
```diff
@@ -22,3 +25,8 @@ def tsig():
 def tsig2d():
 
     yield np.random.randn(2, 1000)
+
+@pytest.fixture(scope='session')
+def tsig_sine():
+
+	yield sim_oscillation(N_SECONDS, FS, freq=FREQ_SINE, variance=None, mean=None)
```
TomDonoghue, line 32: I updated this to use general default test parameters, and so not package and return this data. 

I like this in part because any test using this signal can use the settings value to know things like FS, without needing all the unpacking code.

neurodsp/tests/test_timefrequency_hilbert.py
```diff
@@ -1,42 +1,83 @@
...
 
-def test_amp_by_time(tsig):
+    # Check the expected answer for the test sine wave signal
+    #   The instantaneous phase of sin(t) should be piecewise linear with slope 1
+    phase = phase_by_time(tsig_sine, FS)
 
+    # Create a time axis, scaled to the range of [0, 2pi]
+    times = 2 * np.pi * create_times(N_SECONDS, FS)
+    # Generate the expected instantaneous phase of the given signal
+    answer = np.array([time-np.pi/2 if time <= 3*np.pi/2 else time-5*np.pi/2 for time in times])
```
TomDonoghue, line 55: This line is a bit dense - and to me is a bit of a mystery (at least, takes some time to parse). Could you add some description to give a sense of what this is?

I also think we should consolidate on using the name `expected_answer` or maybe `exp_answer`. Maybe it's just me, but I read `answer` as "proposed answer" (as in, the function return) and not necessarily to answer to check against.

neurodsp/tests/test_timefrequency_hilbert.py
```diff
@@ -1,42 +1,83 @@
...
+def test_amp_by_time(tsig, tsig_sine):
+
+    # Check that a random signal, with a filter applied, runs & preserves shape
     out = amp_by_time(tsig, FS, (8, 12))
     assert out.shape == tsig.shape
 
-def test_freq_by_time(tsig):
+    # Instantaneous amplitude of sinusoid should be 1 for all timepoints
+    amp = amp_by_time(tsig_sine, FS)
+    answer = np.ones_like(amp)
```
TomDonoghue, line 67: You were using a lot of times vectors that I think weren't needed. For these cases `ones_like` seemed like a good option for what was needed.

neurodsp/tests/test_timefrequency_hilbert.py
```diff
@@ -1,42 +1,83 @@
...
+    assert np.max(abs(answer-amp)) < EPS
```
TomDonoghue, line 68: This, and I think all similar asserts, should be using numpy array comparisons (?). 

As in, I think this line could / should be something like:
`assert np.all_close(answer, amp, atol=EPS)`
^unless there's a reason not to use numpy array comparisons, I think we should generally use these.

neurodsp/tests/test_timefrequency_hilbert.py
```diff
@@ -1,42 +1,83 @@
...
 
 from neurodsp.timefrequency.hilbert import *
 
+from neurodsp.utils.data import create_times
+
 ###################################################################################################
 ###################################################################################################
 
-def test_robust_hilbert():
+def test_robust_hilbert(tsig_sine):
```
TomDonoghue, line 14: I didn't really step through and refactor this test func like I did with the phase / amp / freq ones below. Could you check the updates below, and see if there are any analogous updates for this function?

neurodsp/tests/test_filt_filter.py
```diff
@@ -21,6 +23,18 @@ def test_filter_signal(tsig):
     with raises(ValueError):
         out = filter_signal(tsig, FS, 'bandpass', (8, 12), filter_type='bad')
 
+    # Apply lowpass to low-frequency sine. There should be little attenuation.
```
TomDonoghue, line 26: These look like great tests - I just sorta think they are in the wrong place. 

Each test function should test that function specifically. In this case, `filter_signal` is really a pass through function. As set up, these tests test specific FIR filters. So, this means that if this test fails, it would suggest a problem in `filter_signal`, but the actual issue would most likely be in `filter_signal_fir`. So I think these tests should probably be in `test_filter_signal_fir`.

TomDonoghue:
@elybrand - thanks for the updates here! These look much better / more specific tests than we had before. I've make a some comments / suggestions. A bunch of these things are more code mechanics, so if you prefer you can tag in @ryanhammonds  to work on the code and organizational elements here, and we can coordinate this one as a team effort. 

I also wasn't totally sure if this PR was done, or a work-in-progress (if you plan on adding more test updates). Note that if things are in progress, you can use the tag `[WIP]` in the title, so that everyone knows there's more coming.

neurodsp/tests/test_timefrequency_hilbert.py
elybrand: I agree this seems unusual at first. Sin has a period of 2pi seconds, so it's easier to know what the phase should be at time t in [0, 2pi] rather than in [0,1]. The operation `time-np.pi/2 if time <= 3*np.pi/2 else time-5*np.pi/2` in the list comprehension adjusts for the fact that phase is defined to be in [-pi, pi] rather than [0, 2pi].

neurodsp/tests/test_timefrequency_hilbert.py
elybrand: Oh how nice. I was unaware of this function.

neurodsp/tests/test_timefrequency_hilbert.py
elybrand: This is fine with me. I was unaware of this function as well. I imagine it's just a wrapper for what I wrote out in the assert clause but it's an easy fix.

TomDonoghue: Yeh - I don't think it changes the function (I agree it's basically a wrapper on this), but I think it might be a little more readable.

neurodsp/tests/test_timefrequency_hilbert.py
TomDonoghue: Ahh, I see. So, what about adding a `create_radians` function, to add to `utils`, to work a bit like `create_times` and address this use case. This would simplify here, but might be somewhat generally useful to add to the module. 

API would probably be:
`def create_times(n_radians, fs, start_val=0)`

Which could be used in this test as :
`rads = create_times(2*np.pi, fs, start_val=-np.pi)`

neurodsp/tests/test_filt_filter.py
elybrand: Makes sense to me!

neurodsp/tests/test_filt_filter.py
```diff
@@ -2,11 +2,13 @@
 
 from pytest import raises, warns
 
-from neurodsp.tests.settings import FS
+from neurodsp.tests.settings import FS, EPS_FILT
```
TomDonoghue, line 5: now that the tests have moved, the `EPS_FILT` import can be dropped - and the import numpy below.

neurodsp/tests/test_timefrequency_hilbert.py
```diff
@@ -1,42 +1,88 @@
...
+    # Check that a random signal, with a filter applied, runs & preserves shape
     out = freq_by_time(tsig, FS, (8, 12))
     assert out.shape == tsig.shape
 
+    # Instantaneous frequency of sin(t) should be 1 for all timepoints
+    freq = freq_by_time(tsig_sine, FS)
+    expected_answer = np.ones_like(freq)
+
+    assert np.allclose(expected_answer[1:], freq[1:], atol=EPS)
+
```
TomDonoghue, line 85: In your updates, all the funcs get run without a filter, which makes the next function `test_no_filters` not needed. I think we can drop this now extra test - so please just remove the `test_no_filters` function!

neurodsp/tests/test_spectral_power.py
```diff
@@ -40,6 +44,26 @@ def test_compute_spectrum_welch(tsig):
     freqs, spectrum = compute_spectrum_welch(tsig, FS, avg_type='median')
     assert freqs.shape == spectrum.shape
 
+    # Create a sinusoid with 10 periods/cycles.
+    n_seconds = 10
+    fs = 100
+    sig = sim_oscillation(n_seconds, fs, 1, n_cycles=n_seconds, mean=None, variance=None)
```
TomDonoghue, line 50: The `n_cycles` input isn't needed / doesn't do anything here (at least I don't think it does), and shouldn't be needed.

neurodsp/tests/test_spectral_power.py
```diff
@@ -50,5 +74,20 @@ def test_compute_spectrum_wavelet(tsig):
 
 def test_compute_spectrum_medfilt(tsig):
 
+    # Create a sinusoid with 10 periods/cycles.
+    n_seconds = 10
+    fs = 100
+    sig = sim_oscillation(n_seconds, fs, 1, n_cycles=n_seconds, mean=None, variance=None)
```
TomDonoghue, line 80: Same note on `n_cycles`, but also, since we re-use this signal, perhaps we should just add it as a fixture in conftest? It could be `tsig_sine_long` or something,  and use a `N_SECONDS_LONG` settings.

TomDonoghue:
This looks great @elybrand! I've made some small comments for minor organizational things. Once you get a chance to check and update those, then I think this should be all good!

neurodsp/tests/test_spectral_power.py
elybrand: Yup, good call!

TomDonoghue:
Looks good to me, merging in!

PR #206, opened 2020-07-10
[ENH] Conda Support

ryanhammonds:
I'm not a conda user, however, I went ahead and created a script to push pypi builds to conda. Neurodsp may now be install using `conda install -c neurodsp-tools neurodsp`. The shell script should only be execute when a new pypi build is available, although it shouldn't hurt if ran twice on accident, as it defaults to the latest pypi build. The script will prompt for an Anaconda username/password. Accounts may be created [here](https://anaconda.org/). If anyone wants to manage the Anaconda orgs, fell free to send me your Anaconda username, and I can added owners to the Anaconda organization(s).

I have created neurodsp-tools, fooof-tools, bycycle-tools, and lisc-tools organizations on Anaconda. The shell script just has to be adjusted a bit and all these repos will become conda installable.

Edit: Ah, I just noticed the conda issue mentions conda-forge, rather than anaconda. I think the installation command will be the same, it just changes were the conda package is hosted.

### DIFF
--- a/conda/build.sh
+++ b/conda/build.sh
@@ -1,0 +1,36 @@
+#!/usr/bin/bash
+
+# Conda-forge instructions:
+#   - Fork the feedstock repo (https://github.com/conda-forge/neurodsp-feedstock).
+#   - Update the meta.yaml file.
+#   - Make a pull request. Once merged, the package will be rebuilt on the conda-forge channel.
+
+# Anaconda is required to upload to Anaconda Cloud.
+if [[ $(which anaconda) == '' ]]; then
+    echo "Anaconda is not installed or not in \$PATH. Exiting."
+    exit 1
+fi
+
+# Install conda-build if not found.
+if [[ $(conda list | grep conda-build) == '' ]]; then
+    echo "Installing conda-build."
+    conda install conda-build
+fi
+
+# Install anaconda-client if not found.
+if [[ $(conda list | grep anaconda-client) == '' ]]; then
+    echo "Installing anaconda-client."
+    conda install anaconda-client
+fi
+
+# Build
+mkdir -p build
+echo "Building..."
+conda build neurodsp --output-folder=build >> build.log 2>&1
+build=$(conda build neurodsp --output-folder=build --output)
+
+# Upload. Note: this will prompt a username/password
+echo "Uploading to cloud..."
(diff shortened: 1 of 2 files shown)


ryanhammonds:
@TomDonoghue This should be ready for conda-forge. The channel may be specified as `-c neurodsp-tools` or `-c conda-forge`, once the build PR is approved over in the[ conda-forge recipe repo](https://github.com/conda-forge/staged-recipes/pull/12127/).

TomDonoghue:
@ryanhammonds - thanks a ton for doing the dive into conda!

Splitting up my comments by channel:

#### conda-forge

So, I had a quick look at the conda feedstocks, saw that is all set up, and check that I can install `neurodsp` with conda through the `conda-forge` channel. All that seems to be working, and is great! Thanks for setting up. 

My question: do we need the `meta.yaml` file in this repository? 

My understanding is that to update the conda-forge feedstock, we PR and update the yaml on that repository. As far as I can tell, the yaml file in this repository doesn't do anything, and is just likely to fall out of date (listing the wrong version, etc). As far as I see in other conda-forge installable packages this yaml file doesn't seem to need to in the repo. So maybe we can drop it from here?

#### custom channels

So, I think I have a sense of repositories on anaconda, and custom channels, etc (which is an interesting thing that I didn't really know existed), but I'm not too clear on if there is a benefit of having these custom channels as well as using conda-forge.

Also, the `neurodsp-tools` channel doesn't work for me. I think it's because what is available on the neurodsp-tools channel is a build for linux (I'm on mac) and also, only for 3.7. (See available files here: https://anaconda.org/neurodsp-tools/neurodsp/files)

This single build, by itself, doesn't seem all that useful, and I don't know that it's really worth it to try and set things up to make the custom channel work more generally. 

So, my current thought: is there an argument for the custom-channel, and a way to make it more useable? Otherwise, I think we scrap using the custom channel, and stick with using conda-forge.

ryanhammonds:
We can stick with conda-forge since there isn't really any added benefit and it will be easier to maintain from one location. This PR can be closed without merging and we maintain from the conda feedstock.

The meta.yaml was for the custom build, but I messed up the setup since it can only be built in linux. Another reason to stick with conda-forge.

Issue #192, opened 2020-06-20
Organize tests into folders (?)


TomDonoghue:
At some point when  updated FOOOF, I re-organized the tests into folder. I find this makes them a little easier to navigate, and to find things in the tests, etc. Maybe we should update NDSP tests to do the same. 

@ryanhammonds - do you think this would be a useful update?

ryanhammonds:
I think either organization is okay. I'm use to the `test_<folder>_<module>.py ` structure, but a `<folder>/test_<module>.py` may be a bit easier to navigate. I think pytest searches recursively for test_*, so there shouldn't be any pytest setting that need to be changed?

TomDonoghue:
Yeh, it makes no difference to `pytest`, and I don't think any settings need changing. 

I think the current layout is somewhat more common, but I just don't find it scales all that well as the module gets larger, in terms of finding things. It's a bit of file moving, but otherwise a pretty easy update. 

If we do this, I think it can be a 2.2 target, so not a priority for the next release, and we can come back to it for next release.

TomDonoghue:
Done in #207

Issue #191, opened 2020-06-20
Make plot tests save out plots


TomDonoghue:
In FOOOF, some of the plots get saved out in demo-reports and so on. This, I find useful to eyeball these files when something related changes. Otherwise,  plot-tests do a sanity check run, but don't by themselves offer much indication of if the plot runs, but goes weird. 

What if we make all NDSP plot tests save out a copy of the plot to a `test_files` folder. Then, if there is a change to plots, it's easier to go and check these files? This can be modeled on how FOOOF does so (you can see now it sets this up in `conftest`). 

Since we have the `savefig` decorator, any plot call can be saved by adding this to the call line:
`plot_thing(data, save_fig=True, file_name='plot_thing', file_path=TESTS_PLOTS_PATH)`

@ryanhammonds - do you think this is worth adding / updating? I think it would be a fairly small update.

ryanhammonds:
I think this is worth adding. It's more convenient than checking back and forth in a notebook, and like you said shouldn't be too much to implement. I'll work on this and make a PR.

TomDonoghue:
Done in #207

PR #207, opened 2020-07-14
[ENH] Tests reorganization / Test plots saving

ryanhammonds:
This address #191 and #192. Tests files have been reorganized into subfolders to mimic the modules organization (i.e. tests/test_burst_dualthresh.py => tests/burst/test_dualthresh.py) similar to FOOOF. This was done in the first commit. The second commit saves plotting test files to neurodsp/tests/test_files/plots, similar to FOOOF.

Also, neurodsp/tests/plts/test_time_series.py appears as a newfile rather than a rename+modify. I hoped splitting these changes into two commits would resolve this. But git insists that test_times_series.py is a newfile.


TomDonoghue:
This looks great, thanks for the updates @ryanhammonds! Merging in.

PR #208, opened 2020-07-19
[BUG] - Fix the sims normalizations

TomDonoghue:
This PR drops the option in the normalization to only apply transforms to non-zero. This approach, designed to be able to normalize burst regions (ignoring surrounding zeros) turned out to be a bad idea, because there can be other zeros. This created some 'discontinuities' (issue noted in #174). In order to be able to normalize burst signals, without normalizing to non-burst regions, this PR also updates how normalizations get applied in `sim_burst_normalization`, applying them to the simulated signal before tiling. 

Note that this PR replaces #189, which was a wrong-headed fix, that tried to do resampling. 

Review Notes:
@ryanhammonds : if you can review this one, keep in mind it supercedes & adopts from #189, so you can review in light of what you did for that PR. 
@elybrand : since you've played with the sims and normalizations a little, I just wanted to tag you in to sanity check, if you want / can, that this makes sense to you (basically - see if you spot anything mathematically suspect, otherwise don't worry about code-level reviews). 
@rdgao : if you're interested, this is basically the replacement for #189 

Note: the commit log is slightly funky, because I rebased, and I think I didn't do it quite right... but the file diffs seem right.

### DIFF
--- a/neurodsp/sim/periodic.py
+++ b/neurodsp/sim/periodic.py
@@ -2,8 +2,9 @@
 
 import numpy as np
 
-from neurodsp.utils.decorators import normalize
 from neurodsp.sim.transients import sim_cycle
+from neurodsp.utils.norm import normalize_sig
+from neurodsp.utils.decorators import normalize
 
 ###################################################################################################
 ###################################################################################################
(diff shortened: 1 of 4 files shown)


elybrand:
These changes look good to me!

Issue #195, opened 2020-06-22
Tech-Audit #3: Simulations


TomDonoghue:
This issue is part of a 'technical audit' of `neurodsp`, with @elybrand. 

A key goal of neurodsp is to have simple simulations that create example data that exhibits descriptively similar propertties to neural data. 

For the audit of the simulations, I think the key questions are:
- A technical check of the current simulations. Are they, as proposed and implemented, technically appropriate and well defined and described for their stated goals?
- Use this as an opportunity to check and consider if current frameworks & approaches for simulations are well-posed for what we want. Are there things we should extend? Should we consider switching strategy / approach for simulations, and/or adding additional approaches / frameworks? Are there any additional features of real data we would like to simulate that we can add to the simulations?

voytek:
This matches pretty well with the intro email I'd sent, and expands on it nicely, so thanks, @TomDonoghue! The only thing I'd add more explicitly with regards to simulation is to also leverage the simulations for generating ground-truth data for unit testing as well.

elybrand:
I made my first round of looking over the simulations. So far, everything looks pretty good. Here are my thoughts so far.

### Small bugs
There are a few things I caught that we may want to add to the smoke tests. For example, many of the transients have a kwarg which specifies a proportion of time which a rise or fall occurs, but there's no checking if this proportion is strictly between (0, 1). Other kwargs that ask for a decay time don't necessarily restrict the time to be positive. Some examples include

```python
from neurodsp.sim import sim_cycle

n_seconds = 1
fs = 100
sig = sim_cycle(n_seconds, fs, cycle_type='sawtooth', width=1.2) # This seems semantically problematic.
sig = sim_cycle(n_seconds, fs, cycle_type='asine', rdsym=1.01) # This seems semantically problematic.
sig = sim_cycle(n_seconds, fs, cycle_type='exp', tau_d=-1) # Do we want a negative decay time, e.g. exponential blow-up?
```

I'm also a bit confused for the particular aperiodic sim `sim_poisson_pop`. Sure, when the number of neurons is large you expect central limit theorem to give you an approximately Gaussian distribution. But you can directly simulate a poisson distribution, so why not do that?

### Potential additions
The trouble I have with coming up with new sims is that 1) I haven't looked at enough real neuro signals to know what would be reasonable approximations and 2) Anything more complex than sinusoids or dirac spikes make it challenging to test other methods for accuracy.

[...]

Issue #187, opened 2020-06-14
Update simulations to allow for variable start phases


TomDonoghue:
Currently all simulations are simulated with the same starting phase. This is limiting for simulating signals with multiple rhythmic components, as they all end up phase locked. 

ToDo: look into and add a way to simulate rhythmic signals with different start phases. Ideally, this is a general approach that can work across all the different cycles we support.

PR #210, opened 2020-07-20
[ENH] - Update cycle approach

TomDonoghue:
This started as a quick update to cycles, and snowballed a little into some slightly bigger updates / orgs, but I think it all makes sense. 

Changes:
- update the approach for `sim_cycle`, so that it grabs available cycle functions, rather an explicit conditionals
    - to make this approach consistent, each cycle was updated to be it's own function
- separating out cycles from transients
    - this came along as part of the above, and I think is a sensible / useful conceptual organization
- parameter checking of inputs for cycles (related to #195)
    - this lead to some overall updates to checker code
- update to be able to phase shift cycles, to be able to have variable start phases for oscillation sims (#187)

Review notes:
- @ryanhammonds - if you can review for code, as normal.
- @elybrand - I tagged you for checking if this matches what you were thinking in #195 for checking sim params
    - also, could you sanity check the phase shift approach. No mathematical suprises on this?
- @rdgao - if you can sanity check, does the slightly updated organization make sense to you?


neurodsp/sim/cycles.py
```diff
@@ -0,0 +1,275 @@
...
+from scipy.signal import gaussian, sawtooth
+
+from neurodsp.utils.checks import check_param
+from neurodsp.sim.info import get_sim_func
+from neurodsp.sim.transients import sim_synaptic_kernel
+
+###################################################################################################
+###################################################################################################
+
+def sim_cycle(n_seconds, fs, cycle_type, phase=0, **cycle_params):
```
ryanhammonds, line 13: `phase` isn't used in this func.

neurodsp/sim/info.py
```diff
@@ -23,6 +21,8 @@ def get_sim_funcs(module_name):
         A dictionary containing the available sim functions from the requested sub-module.
     """
 
+    from neurodsp.sim import periodic, aperiodic, transients, combined, cycles
```
ryanhammonds, line 24: Why were these imports moved to inside the func?

neurodsp/sim/cycles.py
```diff
@@ -0,0 +1,275 @@
...
+    Parameters
+    ----------
+    n_seconds : float
+        Length of cycle window in seconds.
+        This is NOT the period of the cycle, but the length of the returned array of the cycle.
+    fs : float
+        Sampling frequency of the cycle simulation.
+    cycle_type : {'sine', 'asine', 'sawtooth', 'gaussian', 'exp', '2exp'}
+        What type of cycle to simulate. Options:
+
```
elybrand, line 25: There is no docstring for `phase`.

neurodsp/sim/cycles.py
TomDonoghue: ahh, good catch! At one point I applied phase rotation here, but I think it makes more sense in sim_periodic. I've dropped this argument from here.

neurodsp/sim/info.py
TomDonoghue: Circular imports! All hell breaks loose if they are at the top (it breaks). 

This function is programmatically quirky (not ideal, in best practices sense), but it's efficient at what it does, so I sorta like it.

neurodsp/sim/cycles.py
```diff
@@ -0,0 +1,275 @@
...
+    Returns
+    -------
+    cycle : 1d array
+        Simulated sine cycle.
+
+    Examples
+    --------
+    Simulate a cycle of a 1 Hz sine wave:
+
+    >>> cycle = sim_asine_cycle(n_seconds=1, fs=500)
```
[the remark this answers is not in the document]
TomDonoghue: Ooops, yeh, copy paste errors on the docstrings. Fixed.

neurodsp/sim/cycles.py
TomDonoghue: Ah, phase in this func was accidentally leftover from when I was playing with where to apply phase shifts. I've dropped it from this func, and removed it from the signature.

Issue #214, opened 2020-07-30
Deprecate Hilbert increase N option


TomDonoghue:
When our time-frequency code was originally written, the scipy's Hilbert transform could be really slow for certain lengths 
(see: https://github.com/scipy/scipy/issues/6324)

This issue seems to have been fixed upstream, with scipy adopting updated FFT practices (see the end of the issue and this PR: https://github.com/scipy/scipy/pull/10361). 

This means, I think, that under the assumption that most people are (or should be) using modern scipy versions, that we can deprecate the procedure on our end. This means we can drop the `hilbert_increase_n` argument that is propagated around to control this. 

@elybrand : if you don't mind, could you have a quick look at the linked issue, and check that you have the same impression of it that I do - that our implementation of `robust_hilbert` that does zero-padding is no longer needed?

If so, @ryanhammonds - can you make a PR that removes this option?
Technically this is a breaking change - but I somewhat feel like probably not many people explicitly set this arg, so whole-sale removing it should hopefully not be too disruptive. If we want to be fancy we could add a deprecation warning for 2.2 and totally remove after, but that might be more than is needed...

elybrand:
Yes this looks good to me. I ran the code mentioned in the issue after updating scipy and the difference in runtime between a hilbert transform of a signal with 50,000 entries and with 49,999 entries differs only by a factor of 5 as compared to the factor of 3200 before the fix.

TomDonoghue:
Okay, cool, sounds good! Thanks for checking in @elybrand. This is done in #215

PR #215, opened 2020-07-30
[FIX] Deprecated hilbert_increase_n arg

ryanhammonds:
This is in response to #214, `hilbert_increase_n` and `increase_n` arguments have been removed and tests are updated.


TomDonoghue:
Looks good to me, thank @ryanhammonds! Merging in!

PR #216, opened 2020-08-01
[ENH] - Simulate fractional Gaussian noise and fractional Brownian motion.

elybrand:
This PR adds aperiodic simulations for fractional brownian motion and fractional gaussian noise for a given Hurst parameter in the range (0,1). Effectively this means we have another means of simulating aperiodic signals with a known power law exponent. The simulation of fractional gaussian noise use the Cholesky method for simplicity which runs in O(N^2) flops for a signal of length N. Fractional brownian motion is generated by calculating the cumulative sum of the corresponding fractional gaussian noise. For further details on simulating fractional brownian motion and fractional gaussian noise, [see Ton Dieker's thesis](http://www.columbia.edu/~ad3217/fbm/thesis.pdf), for example.

Below is a minimal working example of simulations for standard brownian motion and positively correlated brownian motion. A simple linear regression is fit to the log-log power spectrum and the relative error of the slopes is compared to what the true values are.

```python
import numpy as np
from neurodsp.plts import plot_time_series, plot_power_spectra
from neurodsp.sim import sim_fbm
from fooof import FOOOF

np.random.seed(0)

n_seconds = 1
fs = 2*10**3

# Simulate standard brownian motion (power law exponent = 2)
# and positively correlated brownian motion (power law exponent = 2.5).
# Recall that the power law exponent and the Hurst exponent are related by
# beta = 2H + 1 for brownian motion.
sig_bm = sim_fbm(n_seconds, fs, hurst=0.5)
sig_pbm = sim_fbm(n_seconds, fs, hurst=0.75)
plot_time_series(times, [sig_bm, sig_pbm], labels=["Brownian Motion", "Positively Correlated Brownian Motion"])

# Plot power spectra
freqs, pspec_bm = compute_spectrum(sig_bm, fs)
_, pspec_pbm = compute_spectrum(sig_pbm, fs)
plot_power_spectra(freqs, [pspec_bm, pspec_pbm], labels=["Brownian Motion", "Positively Correlated Brownian Motion"])

# Use spectral fitting
exp_bm, _ = np.polyfit(np.log10(freqs[1:]), np.log10(pspec_bm[1:]), 1)
exp_pbm, _ = np.polyfit(np.log10(freqs[1:]), np.log10(pspec_pbm[1:]), 1)

rel_err_bm = np.abs(-2 - exp_bm)/2
rel_err_pbm = np.abs(-2.5 - exp_pbm)/2.5

print("Relative error of estimated power law exponent for Brownian Motion: {:.2f}".format(rel_err_bm))
print("Relative error of estimated power law exponent for positively correlated Brownian Motion: {:.2f}".format(rel_err_pbm))
```


neurodsp/sim/aperiodic.py
```diff
@@ -238,6 +239,82 @@ def sim_powerlaw(n_seconds, fs, exponent=-2.0, f_range=None, **filter_kwargs):
...
+def sim_fgn(n_seconds, fs, hurst=0.5):
+    """Simulate a fractional gaussian noise time series with specified Hurst exponent.
+
+    Parameters
+    ----------
+    n_seconds : float
+        Simulation time, in seconds.
+    fs : float
+        Sampling rate of simulated signal, in Hz.
+    hurst : float, optional, default: -0.5
```
TomDonoghue, line 252: There seems to be some discrepancy on whether Hurst exponents are written as positive or negative. 

Here, the value should match the signature, so I think it should be 0.5 (without negative). 

Relatedly, the values in the `Examples sections` seem to use negative, when I think they should also be positive?

neurodsp/sim/aperiodic.py
```diff
@@ -238,6 +239,82 @@ def sim_powerlaw(n_seconds, fs, exponent=-2.0, f_range=None, **filter_kwargs):
...
+
+    Returns
+    -------
+    sig: 1d array
+        Fractional gaussian noise time series with the desired Hurst exponent.
+
+    Notes
+    -----
+    The relationship between the power law exponent beta and the Hurst exponent for fractional
+    gaussian noise is hurst = (beta + 1)/2.
```
TomDonoghue, line 263: we otherwise don't use 'beta' to refer to what we generally just call the 'aperiodic exponent'. In neuro / in FOOOF, it's often referenced as 1/f^x, or as `/chi`. How do we feel about calling it `chi` here?

neurodsp/sim/aperiodic.py
```diff
@@ -238,6 +239,82 @@ def sim_powerlaw(n_seconds, fs, exponent=-2.0, f_range=None, **filter_kwargs):
...
+    >>> sig = sim_fgn(n_seconds=1, fs=500, hurst=0.5)
+    """
+
+    # Construct the autocovariance function.
+    n_samples = int(n_seconds * fs)
+    gamma = np.zeros(n_samples)
+    for k in range(n_samples):
+        gamma[k] = 0.5*(np.abs(k-1)**(2 * hurst) - 2*k**(2*hurst) + (k+1)**(2*hurst))
+
+    # Build the autocovariance matix. Use the Cholesky factor to transform white noise to get the desired time series.
```
TomDonoghue, line 278: typo on `matix`

neurodsp/sim/aperiodic.py
```diff
@@ -238,6 +239,82 @@ def sim_powerlaw(n_seconds, fs, exponent=-2.0, f_range=None, **filter_kwargs):
...
+    gaussian noise is hurst = (beta + 1)/2.
+
+    Examples
+    --------
+    Simulate fractional gaussian noise with an exponent of -0.5 (white noise):
+
+    >>> sig = sim_fgn(n_seconds=1, fs=500, hurst=0.5)
+    """
+
+    # Construct the autocovariance function.
```
TomDonoghue, line 272: this is pure nitpicking - but by convention we don't add periods to inline comments (unless there are multiple sentences)

neurodsp/sim/aperiodic.py
```diff
@@ -238,6 +239,82 @@ def sim_powerlaw(n_seconds, fs, exponent=-2.0, f_range=None, **filter_kwargs):
...
+        Fractional brownian motion with the desired Hurst exponent.
+
+    Notes
+    -----
+    The relationship between the power law exponent beta and the Hurst exponent for fractional
+    brownian motion is hurst = (beta - 1)/2.
+
+    Examples
+    --------
+    Simulate fractional brownian motion with an exponent of -0.5 (brown noise):
```
TomDonoghue, line 310: Since we're introducing a new exponent, I think we should be explicit which exponent we are talking about, since in general we talk about the 'aperiodic exponent' as shorthand for 'powerlaw exponent' - and this confused me for a minute 

tl-dr: I think we explicitly say `Hurst exponent of` here (and check to explicitly note when we mean Hurst exponent throughout).

neurodsp/sim/aperiodic.py
```diff
@@ -238,6 +239,82 @@ def sim_powerlaw(n_seconds, fs, exponent=-2.0, f_range=None, **filter_kwargs):
...
+
+    >>> sig = sim_fbm(n_seconds=1, fs=500, hurst=0.5)
+    """
+
+    # Fractional brownian motion is just the cumulative sum of fractional gaussian noise.
```
TomDonoghue, line 315: nitpick, but doesn't need the word `just`

neurodsp/tests/sim/test_aperiodic.py
```diff
@@ -33,6 +35,26 @@ def test_sim_powerlaw():
     sig = sim_powerlaw(N_SECONDS, FS, f_range=(2, None))
     check_sim_output(sig)
 
+def test_sim_fgn():
+
+    # Simulate white noise.
+    sig = sim_fgn(N_SECONDS, FS)
+    check_sim_output(sig)
+
+    # Check the accuracy of the mean and standard deviation.
```
TomDonoghue, line 44: It seems to me that these check the performance of the normalization, and not the simulation per se? If this is the case (if I'm not not missing something about it), I don't think we want these checks, as tests should be modular, directly testing only the function under question, and this tests something else (alternately stated: if this test fails, it's not `sim_fgn` that is broken).

neurodsp/tests/sim/test_aperiodic.py
```diff
@@ -33,6 +35,26 @@ def test_sim_powerlaw():
...
+    np.allclose(np.mean(sig), 0, atol=0.1)
+    np.allclose(np.std(sig), 1, atol=0.1)
+
+def test_sim_fbm():
+
+    # Simulate standard brownian motion.
+    sig = sim_fbm(N_SECONDS, FS)
+    check_sim_output(sig)
```
TomDonoghue, line 52: to add an accuracy check, we can add a check that the hurst exponent is as expected, using the now merged fluctuations function, right?

TomDonoghue:
I have pushed a refactor of the auto-covariance estimation, as that step was a bit slow (because of the explicit loop). 
I played with lambda's / vectorizing, to get it to work, and I can get an ~50x speed-up. 

I don't think I broke anything in the process. Some profiling coped below, with example outputs. For every version I tried here, the two get the same answer. @ryanhammonds - how familiar are you with vectorizing? Does this look like a good update to you?

Refactor checks:
```
# Setup code
import numpy as np
n_seconds = 10
fs = 1000
hurst = 0.75
```

```
%%timeit
n_samples = int(n_seconds * fs)
gamma1 = np.zeros(n_samples)
for k in range(n_samples):
    gamma1[k] = 0.5*(np.abs(k-1)**(2 * hurst) - 2*k**(2*hurst) + (k+1)**(2*hurst))
```
> 7.34 ms ± 111 µs per loop

```
%%timeit
n_samples = int(n_seconds * fs)
gamma2 = np.arange(0, n_samples)
def autocov(hurst):
    return lambda k : 0.5*(np.abs(k-1)**(2 * hurst) - 2*k**(2*hurst) + (k+1)**(2*hurst))
gamma2 = np.apply_along_axis(autocov(hurst), 0, gamma2)
```
> 157 µs ± 428 ns per loop

TomDonoghue:
Thanks @elybrand for the PR - this looks awesome! Thanks for the code example, that works for me. 

I reviewed the code directly and left some direct comments, mostly on documenting, nitpicks etc. 

A couple more general thoughts:
- I wonder if there are options for slight more clear function names?
    - The current names seem a little opaque, but I'm not sure what good alternatives are, that aren't too long
- I know this differs from a lot math code, but we generally avoid single character or capitalized variable names (for PEP8 compliance, or in my mind 'every variable should be control-f-able'). How do you feel about updating style of the var names?

One thing I'm a little unclear on is the convention for Hurst exponent as positive or negative. I also tried our new function that computes Hurst exponent, and it seems the computed Hurst for `sim_fbm` is 1 minus the input Hurst. Is this expected?
```
from neurodsp.aperiodic import compute_fluctuations

n_seconds = 10
fs = 1000
hurst = 0.8

white = sim_fgn(n_seconds, fs, hurst=hurst)
brown = sim_fbm(n_seconds, fs, hurst=hurst)

_, _, h_white = compute_fluctuations(white, fs, method='rs')
_, _, h_brown = compute_fluctuations(brown, fs, method='rs')

print(h_white)
print(h_brown)
```
> 0.6954255235498658
> 0.33349871245995044

neurodsp/sim/aperiodic.py
elybrand: Yes, thank you. The Hurst exponent should be positive. This is a typo.

neurodsp/tests/sim/test_aperiodic.py
elybrand: This is a good point. Perhaps it's best to remove the accuracy test. Effectively I was just using a sanity check that the first two moments of the empirical signal agreed with that of a standard gaussian.

elybrand: Or perhaps I avoid using normalization and check the first four empirical moments.

neurodsp/tests/sim/test_aperiodic.py
elybrand: Perhaps. If such a test were to fail, who would be the culprit? The fluctuations code or the fractional brownian motion code? The accuracy tests I used below these lines were a sanity check on the increments being gaussian by checking the first two moments. In the newest commit I do not normalize the signal and check the first four moments as an accuracy test.

elybrand:
@TomDonoghue Thanks for the detailed comments. With regards to the 1-chi phenomenon in your minimal working example, this is as expected. The relationship between the DFA exponent and the power law exponent is indeed 1-chi. [See the wikipedia page for this relation.](https://en.wikipedia.org/wiki/Detrended_fluctuation_analysis#Relations_to_other_methods)

elybrand:
@TomDonoghue Thanks for the detailed comments. I've updated the comments in the code and clarified when we are mentioning the Hurst exponent vs the power law exponent chi.

`I also tried our new function that computes Hurst exponents on the new simulations. It seems to work as it should, except that the computed Hurst for sim_fbm is 1 minus the input Hurst. Is this expected?`

So I don't know the mapping from the rescaled range fluctuations to the power law exponent. However, I do know the mapping under DFA. For DFA, the mapping from the DFA exponent to the power law exponent chi is `chi = alpha` for fractional gaussian noise and `chi = alpha + 1` for fractional brownian motion. The extra + 1 in fractional brownian motion comes from taking the cumulative sum (i.e. integrating). When I run your code with `method = 'dfa'`, I get what I expect:

```python
from neurodsp.aperiodic import compute_fluctuations

n_seconds = 5
fs = 500
hurst = 0.7

white = sim_fgn(n_seconds, fs, hurst=hurst)
brown = sim_fbm(n_seconds, fs, hurst=hurst)

_, _, h_white = compute_fluctuations(white, fs, method='rs')
_, _, h_brown = compute_fluctuations(brown, fs, method='rs')

print(h_white)
print(h_brown)
```
0.7050537065447817
1.6446752365857265

elybrand:
Okay, I changed the naming of the functions. They're about 23 and 24 characters long which is just a few characters longer than `sim_synaptic_current`. 

@TomDonoghue I believe after some digging I found that the mapping from rescaled range exponent `alpha` to the power law exponent `chi` is `alpha = 1 - chi/2`. From your two estimates that your code provides, those should give power law exponent estimates 0.6954255235498658 and 0.8332506437700248, respectively.

elybrand:
Okay, so this has been a mind boggling journey. There appears to be a ton of confusion as to what's going on with all of these methods and under what circumstances they work. Because of this deep dive, I've made the following changes to the code:

1. I've added chi as an explicit kwarg for both simulations. This way, the control over the power law exponent in the spectrogram is explicit. The hurst parameter, if defined, overwrites chi.
2. I've added more verbose documentation explaining that the hurst kwarg.
3. I added value checking for chi and hurst for both sims.

Here is a new minimal working example using the chi kwarg.

```python
import numpy as np
from neurodsp.plts import plot_time_series, plot_power_spectra
from neurodsp.sim import *

np.random.seed(0)

n_seconds = 1
fs = 2*10**2

sig_bm = sim_frac_brownian_motion(n_seconds, fs, chi=2, mean=None, variance=None)
sig_pbm = sim_frac_brownian_motion(n_seconds, fs, chi=2.9, mean=None, variance=None)
sig_nbm = sim_frac_brownian_motion(n_seconds, fs, chi=1.1, mean=None, variance=None)
plot_time_series(times, [sig_nbm, sig_bm, sig_pbm], labels=[ r"$\chi=1.1$", r"$\chi=2$", r"$\chi=2.9$"])

# Plot power spectra
freqs, pspec_bm = compute_spectrum(sig_bm, fs)
_, pspec_pbm = compute_spectrum(sig_pbm, fs)
_, pspec_nbm = compute_spectrum(sig_nbm, fs)
plot_power_spectra(freqs, [pspec_nbm, pspec_bm, pspec_pbm], labels=[r"$\chi=1.1$", r"$\chi=2$", r"$\chi=2.9$"])
```

I think going forward with the aperiodic methods work we should add documentation to each method indicating under what circumstances they should be applied. For anyone who is curious what I found in the deep dive, here is a brief summary.

### The Literature Says Beware the Context
[The methods paper by Brach and Sejdic](https://www.researchgate.net/profile/Ervin_Sejdic/publication/258347281_A_comparative_analysis_of_spectral_exponent_estimation_techniques_for_1f_processes_with_applications_to_the_analysis_of_stride_interval_time_series/links/5a5f8570aca2727352436f63/A-comparative-analysis-of-spectral-exponent-estimation-techniques-for-1-f-processes-with-applications-to-the-analysis-of-stride-interval-time-series.pdf) make it a point to stress that the choice of spectral exponent estimation method depends on whether the time series is stationary (i.e. fractional gaussian noise in our setting) or not (fractional brownian motion). Inappropriate use can lead to misinterpretation of the estimated exponent. 

While Brach and Sejdic make no mention of rescaled range analysis, the authoritative source that they and many other papers I found cite is a paper by [Eke, Herman, Kocsis, and Kozak](https://iopscience.iop.org/article/10.1088/0967-3334/23/1/201/pdf?casa_token=3sVkc2O6XCMAAAAA:owd4OEgoo9pbu-H6P1wqrKNpDfBywgUCh2h2coq0hPBleTwa-rpEIxuv1DBiVrEq0uPBp6Us). In that work, rescaled range analysis is remarked to work on fractional gaussian noise, or **differenced** fractional brownian motion. Unfortunately, the remark is not accompanied by any reasoning to explain why, or a citation that might provide insight into this.

### The Devil is in the Exponent
[Recall that the Hurst EXPONENT (He) is **defined** in the context of rescaled range analyses](https://en.wikipedia.org/wiki/Hurst_exponent)

E[ R_n/S_n ] = pn^{He}

where p is a constant of proportionality and n is the length of the rescaled range segment. On the other hand,[ self-similar time series X(t) are defined by a Hurst INDEX or Hurst PARAMETER  (H)](https://en.wikipedia.org/wiki/Fractional_Brownian_motion) so that

X(at) = a^(H)X(t)

where the equality holds in distribution. This definition is notably absent in the exposition of the Brach and Sejdic paper. The simulation code takes as input a Hurst index, not a Hurst exponent.

# Hurst Exponent != Hurst Parameter 
There doesn't appear to be an immediate mapping from He to H, but the literature indicates that for fractional Gaussian noise He is approximately H. The minimal working example @TomDonoghue provided above illustrates this. It has been empirically seen (even in the above numerical examples) that He is not H when the signal is non-stationary, or fractional brownian motion. It is curious that we are seeing something like H  ~ 1 - He. A relation like this does exist between the two hurst indices of fractional brownian motion (HfBm) and fractional gaussian noise (HfGn), namely 

HfBm + 1  = HfGn,

though it's not immediately clear to me if this has any relevance to the phenomenon we're seeing.

TL;DR: Rescaled range analysis does not directly estimate the Hurst index. It estimates a separate quantity known as the Hurst exponent whose relation to the Hurst index is not always clear.

neurodsp/tests/sim/test_aperiodic.py
```diff
@@ -33,6 +36,30 @@ def test_sim_powerlaw():
     sig = sim_powerlaw(N_SECONDS, FS, f_range=(2, None))
     check_sim_output(sig)
 
+def test_sim_frac_gaussian_noise():
+
+    # Simulate white noise. Do not normalize.
+    np.random.seed(0)
```
ryanhammonds, line 42: I think setting a seed can be removed. If the tests fails because of random variation maybe a wider tolerance can be used.

neurodsp/tests/sim/test_aperiodic.py
```diff
@@ -33,6 +36,30 @@ def test_sim_powerlaw():
...
+    # Check the accuracy of the mean and standard deviation
+    np.allclose(np.mean(sig), 0, atol=0.01)
+    np.allclose(np.std(sig), 1, atol=0.01)
+    np.allclose(skew(sig), 0, atol=0.01)
+    np.allclose(kurtosis(sig), 3, atol=0.01)
+
+def test_sim_frac_brownian_motion():
+
+    # Simulate standard brownian motion. Do not normalize.
+    np.random.seed(0)
```
ryanhammonds, line 54: Same as above.

elybrand:
Hey everyone, I wanted to point out that there is a big caveat to simulating fractional brownian motion for the purposes of spectral estimation. I noticed that when `chi` is near 1, there is an incredibly large bias in estimating `chi` from the fractional brownian motion spectrum. I've spent the past couple of hours making sure there wasn't some kind of bug in the code. 

Here is a minimal working example to show you what's going on
```python
import numpy as np
from scipy.optimize import curve_fit
from neurodsp.sim import sim_combined
from neurodsp.sim import sim_frac_brownian_motion, sim_frac_gaussian_noise
import pandas as pd

def fit_func(xdata, slope, offset):
	return slope*xdata + offset

fs = 1000
n_seconds = 1

# Chi is in the range (1,3) for fractional brownian motion. The smaller chi is
# the worse the bias in estimating it.
chi = 1.01

num_trials = 100

df = pd.DataFrame(data=np.zeros((num_trials, 3)), columns=["trial", "fBm_err", "rotation_err"])
df['trial'] = np.arange(num_trials)

for trial in range(num_trials):
	sig_fbm = sim_frac_brownian_motion(n_seconds, fs+1, chi)

	sim_components = {'sim_powerlaw': {'exponent' : -chi}}
	sig_rot = sim_combined(n_seconds, fs, components=sim_components)

	freqs = np.linspace(1, fs//2, num=fs//2)
	psd_fbm = np.abs(np.fft.fft(sig_fbm)[1:fs//2+1])**2
	psd_rot = np.abs(np.fft.fft(sig_rot)[1:fs//2+1])**2

	[slope_fbm, offset1], _ = curve_fit(fit_func, np.log10(freqs), np.log10(psd_fbm))
	[slope_rot, offset2], _ = curve_fit(fit_func, np.log10(freqs), np.log10(psd_rot))

	fbm_errs = np.log10(10**offset1 * freqs**slope_fbm) - np.log10(psd_fbm)
	rot_errs = np.log10(10**offset2 * freqs**slope_rot) - np.log10(psd_rot)

	df.loc[df['trial'] == trial, 'rotation_err'] = -slope_rot - chi
	df.loc[df['trial'] == trial, 'fBm_err'] = -slope_fbm  - chi

df[['rotation_err', 'fBm_err']].describe()
# fBm bias is on the order of 1
```

It looks like what's happening is that the cumulative summing of fractional gaussian noise is distorting the high frequencies in the simulated fractional brownian motion. Because those signals with `chi` near 1 feature more power in the high frequency range, this may explain why estimating chi for these signals is really bad. 

To make sure the cumulative summing is really the issue, I ran the same experiment for fractional gaussian noise. Here is that code. Notice that the bias for estimating `chi` is significantly reduced.

```python
import numpy as np
from scipy.optimize import curve_fit
from neurodsp.sim import sim_combined
from neurodsp.sim import sim_frac_brownian_motion, sim_frac_gaussian_noise
import pandas as pd

def fit_func(xdata, slope, offset):
	return slope*xdata + offset

fs = 1000
n_seconds = 1

# Chi is in the range (-1, 1) for fractional gaussian motion.
chi = 0.99

num_trials = 100

df = pd.DataFrame(data=np.zeros((num_trials, 3)), columns=["trial", "fGn_err", "rotation_err"])
df['trial'] = np.arange(num_trials)

for trial in range(num_trials):
	sig_fgn = sim_frac_gaussian_noise(n_seconds, fs+1, chi)

	sim_components = {'sim_powerlaw': {'exponent' : -chi}}
	sig_rot = sim_combined(n_seconds, fs, components=sim_components)

	freqs = np.linspace(1, fs//2, num=fs//2)
	psd_fgn = np.abs(np.fft.fft(sig_fgn)[1:fs//2+1])**2
	psd_rot = np.abs(np.fft.fft(sig_rot)[1:fs//2+1])**2

	[slope_fgn, offset1], _ = curve_fit(fit_func, np.log10(freqs), np.log10(psd_fgn))
	[slope_rot, offset2], _ = curve_fit(fit_func, np.log10(freqs), np.log10(psd_rot))

	fgn_errs = np.log10(10**offset1 * freqs**slope_fgn) - np.log10(psd_fgn)
	rot_errs = np.log10(10**offset2 * freqs**slope_rot) - np.log10(psd_rot)

	df.loc[df['trial'] == trial, 'rotation_err'] = -slope_rot - chi
	df.loc[df['trial'] == trial, 'fGn_err'] = -slope_fgn  - chi

df[['rotation_err', 'fGn_err']].describe()
# fGn bias is on the order of 0.08
```

TomDonoghue:
As a note, we had talked about how this simulations create broadly equivalent data to spectral rotation. I still think it's probably worth adding this sims, to be able to compare and use them, if/when needed. 

In terms of quirk with the fractional brownian motion - thanks for the code example @elybrand - it all ran nicely for me. 
As a note, it is a systematic directional bias, as in - it's not high average absolute error, but that (with low input values for chi), the resulting spectrum is always steeper than expected. Representative example:
Screen Shot 2020-09-28 at 3 15 59 PM

Just as a crude visual note, it's not so clear there is a quirk at any particular frequencies, but rather just the whole spectrum is very similar to the spectral-rotation version, but with an unexpected exponent. I don't know why. This is something we want to check / figure out a bit / document - because as it is, this sim can't really be used to create data with the expected chi, across the range we'd like to do so.

ryanhammonds:
I reviewed this thread and explored the code a bit more. The latest commit I pushed fixed an bug where `hurst` was being passed as `chi` in `sim_frac_gaussian_noise`. Running this bit of code pre/post that commit highlights the bug:
```python
import pandas as pd
import numpy as np
from scipy.optimize import curve_fit
from neurodsp.sim import sim_frac_brownian_motion, sim_frac_gaussian_noise

fs = 1000
n_seconds = 10
chis = np.array([2.9, 2.5, 2, 1.5, 1.1, 1.01])

df = pd.DataFrame()
df['chi'] = chis

def _estimate_exponent(xs, offset, chi):
    return xs*chi + offset

df['chi_hat'] = np.zeros_like(chis)
df['error'] = np.zeros_like(chis)

for idx, chi in enumerate(chis):
    
    sig = sim_frac_brownian_motion(n_seconds, fs, chi=chi)
    freqs = np.linspace(1, fs//2, num=fs//2)
    powers = np.abs(np.fft.fft(sig)[1:fs//2+1])**2
    
    ap_params, _ = curve_fit(_estimate_exponent, np.log10(freqs), np.log10(powers), maxfev=5000)
    _, chi_hat = ap_params[:]
    
    df['chi_hat'].iloc[idx] = -chi_hat
    df['error'].iloc[idx] = abs(-chi_hat - chi)
    
```
Before:
```
 chi | chi_hat  | error
--------------------------
2.90 | 2.952483 | 0.052483
2.50 | 2.760382 | 0.260382
2.00 | 2.512107 | 0.512107
1.50 | 2.170583 | 0.670583
1.10 | 2.059431 | 0.959431
1.01 | 1.961432 | 0.951432
```
After:
```
 chi | chi_hat  | error
--------------------------
2.90 | 2.903547 | 0.003547
2.50 | 2.509183 | 0.009183
2.00 | 1.916631 | 0.083369
1.50 | 1.505829 | 0.005829
1.10 | 1.005183 | 0.094817
1.01 | 0.468352 | 0.541648
```

elybrand:
@ryanhammonds eagle eye! This makes me very happy! Not that an error of 0.5 is great, but at least the relative error is no longer on the order of 1 : )

ryanhammonds:
@elybrand Glad I could help! Another part of decreasing error was increasing `n_seconds` from `1` to `10`. For comparison: 

1 second:
```
 chi | chi_hat  | error
--------------------------
2.90 | 2.835745 | 0.064255
2.50 | 2.356110 | 0.143890
2.00 | 1.852651 | 0.147349
1.50 | 1.202260 | 0.297740
1.10 | 0.456141 | 0.643859
1.01 | 0.140335 | 0.869665
```

I feel like I'm still missing something but haven't been able to figure it out.

elybrand:
@ryanhammonds I'd say, barring any other egregious bugs, don't lose any sleep over it. I think if you were to calculate the average periodogram of multiple simulations of fractional brownian motion you *might* see a slightly better error rate. In general, I believe there is a bias in the empirical realizations of the periodogram for discrete fractional noise processes, particularly for anticorrelated time series (i.e. when H is near zero). If you're curious, [here's an article that discusses what's going on](https://www.ncbi.nlm.nih.gov/pmc/articles/PMC3377491/). It appears that there are biases in the periodogram even when H > 0.5, but in that setting the bias is only evident at the high frequencies, which is why I think we're seeing better error rates for larger values of H.

ryanhammonds:
@elybrand Thanks for the explanation and link!

The last two commits flipped the sign of chi and added accuracy tests. I forgot about the conversation above suggesting to use `compute_fluctuations`. However, I couldn't get stable values using `compute_fluctuations` without using a seed:

```python
N_SECONDS = 1
FS_HIGH = 1000

hurst = .5
h_browns = np.zeros(1000)

for i in range(1000):

    sig = sim_frac_brownian_motion(N_SECONDS, FS_HIGH, hurst=hurst)
    
    _, _, h_brown = compute_fluctuations(sig, fs, method='dfa')
    
    h_browns[i] = h_brown - 1
```
```
np.mean(h_browns)
```
> 0.367603202421886
```
np.std(h_browns)
```
> 0.10822828318413678
```
np.min(h_browns)
```
> -0.053093281673274095

TomDonoghue:
ToDo: add to notes - estimation error increases with lower chi values and shorter signals.

TomDonoghue:
(finally) merging this one in! Thanks team!

PR #217, opened 2020-08-02
[ENH] - Add autocorrelation measure

TomDonoghue:
Add a function to compute autocorrelation of time series. 

Note: this implementation was shamelessly taken and adapted from @rdgao's implementation. I'm suggesting we add this functionality, for a stable implementation, including for use in the AperiodicMethods project. Does that sound good to use @rdgao?

For reviewing, the usual: @elybrand please check if the math works / makes sense, and @ryanhammonds please sanity check the code, tests, etc.

### DIFF
--- a/doc/api.rst
+++ b/doc/api.rst
@@ -195,6 +195,7 @@ Fluctuation Analyses
 .. autosummary::
     :toctree: generated/
 
+    compute_autocorr
     compute_fluctuations
     compute_irasa
     fit_irasa
--- a/neurodsp/aperiodic/__init__.py
+++ b/neurodsp/aperiodic/__init__.py
@@ -1,4 +1,5 @@
 """"Functions for calculating aperiodic properties of signals."""
 
 from .dfa import compute_fluctuations
+from .autocorr import compute_autocorr
 from .irasa import compute_irasa, fit_irasa
--- a/neurodsp/aperiodic/autocorr.py
+++ b/neurodsp/aperiodic/autocorr.py
@@ -12,10 +12,10 @@ def compute_autocorr(sig, max_lag=1000, lag_step=1):
     ----------
     sig : array 1D
         Time series to compute autocorrelation over.
-    max_lag : int (default=1000)
+    max_lag : int, optional, default: 1000
         Maximum delay to compute autocorrelations for, in samples.
-    lag_step : int (default=1)
-        Step size (lag advance) for computing correlations.
+    lag_step : int, optional, default: 1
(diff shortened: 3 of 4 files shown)


rdgao:
I did a brief test at some point, comparing two of my implementations with `acf` from `statstool` (https://www.statsmodels.org/devel/generated/statsmodels.tsa.stattools.acf.html), the latter is almost always faster, and by a lot. I don't know how they're doing it, but I'm pretty sure I'm doing it inefficiently, so I'd just import that instead

elybrand:
@TomDonoghue Everything looks good. 

@rdgao They are likely using the FFT to get speed ups. Since the autocovariance function is just a convolution of a signal with its reflection, everything is just pointwise multiplication in the Fourier domain.

rdgao:
@elybrand yeah there is an arg to use FFT, probably for speed up with long data, but I don't think that's the default. I just meant if I'm doing some stupid loop thing where you can vectorize.

@ryanhammonds was that implementation above not horrendously slow compared to statsmodel.acf? Computing the full convolution and then throwing out most of it (since we're typically interested in just a few seconds worth of lag) feels inefficient to me, though convolve uses FT under the hood as well... in any case, if there is no crazy difference in speed, then I agree, those 3 lines is prob better than to introduce a nonstandard dependency

ryanhammonds:
@rdgao It's almost the same implementation as statsmodels.acf and it's about the same speed: 

```
np.random.seed(0)
sig = np.random.randn(1000)
```
The update from my comment:
```
%%timeit
timepoints, autocorrs = compute_autocorr_modified(sig, max_lag=200)
```
> 115 µs ± 107 ns per loop (mean ± std. dev. of 7 runs, 10000 loops each)

statsmodel's:
```
%%timeit
autocorrs = acf(sig, nlags=199, fft=None)
```
> 139 µs ± 175 ns per loop (mean ± std. dev. of 7 runs, 10000 loops each)

The implementation Tom pushed:
```
%%timeit
timepoints, autocorrs = compute_autocorr(sig, max_lag=200, lag_step=1)
```
> 3.52 ms ± 25.8 µs per loop (mean ± std. dev. of 7 runs, 100 loops each)

rdgao:
yep, that checks out - whatever I was doing was 10x slower. good to push for me

TomDonoghue:
Awesome, thanks a lot @ryanhammonds - I think adapting the statsmodels implementation to add this functionality without adding a new dependency was absolutely the right call here - thanks for going ahead with that!

LGTM - merging in.

PR #221, opened 2020-08-31
[ENH] - Simulation of Central Frequency in the Time Domain

elybrand:
## Overview

This PR builds a new combined aperiodic and periodic simulation which allows the user to simulate an aperiodic component with a specified power law exponent and a single oscillatory component with specified central frequency, bandwidth, and relative height above the aperiodic component in the log-power spectrum. 

Here is a minimal working example with the analytical power spectrum superimposed for comparison.

```python
import numpy as np
from neurodsp.plts import plot_power_spectra, plot_time_series
from neurodsp.sim import sim_central_freq

np.random.seed(0)

n_seconds = 10
fs = 10**3
chi = -2
central_freq = 20
bw = 5
ht = 2
sig = sim_central_freq(n_seconds, fs, chi, central_freq, bw, ht)
freqs, psd = compute_spectrum(sig, fs, )#nperseg=fs*n_seconds)

# Build a reference power law to visually track if the psd has exponent chi.
psd_baseline = psd[1]/(freqs[1]**chi)*np.array([10**(ht * np.exp(-(freq - central_freq)**2/(2*bw**2))) * freq**(chi) for freq in freqs])

fig, axes = plt.subplots(2,1, figsize=(15,10))
axes[0].set_title("Aperiodic Noise with Central Frequency")
plot_time_series(times, sig, ax=axes[0])
plot_power_spectra(freqs, [psd, psd_baseline], ax=axes[1], labels=["Empirical PSD", "Reference PSD"])
plt.tight_layout()
```
image

## Comments
The way the code is written is a bit limiting. For example, it doesn't allow the presence of multiple central frequencies. The code is easily adaptable to handle this case though. In general, given a signal in the time domain, this code adds sinusoids at the relevant frequencies to add a ``bump'' in the log power spectrum to the inputted signal. To get multiple central frequencies, you'd just do a few recursive calls on aperiodic noise with different central frequencies, bandwidths, and relative heights.

@ryanhammonds @TomDonoghue Do you think I should rewrite this code so that it takes as input a general time series and then adds sinusoids to augment its power spectrum, or do you think that should be a separate function and we can keep this sim as is?

## The Math Behind The Code
image

### DIFF
--- a/neurodsp/sim/combined.py
+++ b/neurodsp/sim/combined.py
@@ -85,12 +85,14 @@ def sim_combined(n_seconds, fs, components, component_variances=1):
 
     return sig
 
+
 @normalize
-def sim_central_freq(n_seconds, fs, chi, central_freq, bw, ht):
+def sim_central_freq(n_seconds, fs, central_freq, bw, ht, ap_func='sim_powerlaw',
+                     ap_kwargs={'exponent': -2}):
     """
-    Returns a time series whose full power spectrum consists of a power law with exponent chi
+    Returns a time series whose full power spectrum consists of an aperiodic component
     and a gaussian peak at central_freq with standard deviation bw and relative height ht.
-    
+
 
     Parameters
     -----------
@@ -98,14 +100,17 @@ def sim_central_freq(n_seconds, fs, chi, central_freq, bw, ht):
         Number of seconds elapsed in the time series.
     fs: float
         Sampling rate.
-    chi: float
-        Power law exponent.
     central_freq: float
         Central frequency for the gaussian peak in Hz.
     bw: float
         Bandwidth, or standard deviation, of gaussian peak in Hz.
     ht: float
-        Relative height in log_10(Hz) over the aperiodic component of the gaussian peak at central_freq.
+        Relative height in log_10(Hz) over the aperiodic component of the gaussian peak at
+        central_freq.
+    ap_func : str, optional, default: 'sim_powerlaw'
+        An aperiodic simulation function from ``neurodsp.sim.aperiodic``.
+    ap_kwargs : dict, optional, default: {'exponent': -2}
+        Keyword arguments to use with ``ap_func``.
 
     Returns
     -------
@@ -114,17 +119,18 @@ def sim_central_freq(n_seconds, fs, chi, central_freq, bw, ht):
 
     Examples
     --------
-    Simulate aperiodic noise with exponent -2 superimposed over an oscillatory component with
-    central frequency 20.
+    Simulate an aperiodic component with exponent -2 superimposed over
+    an oscillatory component with central frequency 20.
 
-    >>> sig = sim_gauss_peak(n_seconds=50, fs=500, chi=-2, central_freq=20, bw=5, ht=7)
+    >>> sig = sim_central_freq(n_seconds=50, fs=500, central_freq=20, bw=5, ht=7,
+    ...                        ap_func='sim_powerlaw', ap_params={'exponent': -2})
     """
 
     times = create_times(n_seconds, fs)
 
     # Construct the aperiodic component and compute its Fourier transform. Only use the
     # first half of the frequencies from the Fourier transform since the signal is real.
-    sig_components = {'sim_powerlaw': {'exponent': chi}}
+    sig_components = {ap_func: ap_kwargs}
     sig_ap = sim_combined(n_seconds, fs, sig_components)
     sig_len = sig_ap.shape[0]
 
@@ -147,17 +153,19 @@ def sim_central_freq(n_seconds, fs, chi, central_freq, bw, ht):
 
     # Build an array of the amplitude coefficients
     cosine_coeffs = np.array([
-                    (-np.real(sig_ap_hat[ell]) + np.sqrt(np.real(sig_ap_hat[ell])**2 + (10**rel_heights[ell] - 1)*np.abs(sig_ap_hat[ell])**2))/cosine_norms[ell]
+                    (-np.real(sig_ap_hat[ell]) + np.sqrt(np.real(sig_ap_hat[ell])**2 + \
+                        (10**rel_heights[ell] - 1)*np.abs(sig_ap_hat[ell])**2))/cosine_norms[ell]
                     for ell in range(cosine_norms.shape[0])]
                 )
 
     # Add cosines with the respective coefficients and with a random phase shift for each one
     sig_periodic = np.sum(
                     np.array(
-                        [cosine_coeffs[ell]*np.cos(2*np.pi*freqs[ell]*times + 2*np.pi*np.random.rand()) for ell in range(cosine_norms.shape[0])]
+                        [cosine_coeffs[ell]*np.cos(2*np.pi*freqs[ell]*times + \
+                            2*np.pi*np.random.rand()) for ell in range(cosine_norms.shape[0])]
                     ),
                     axis=0
-                    )
+                )
 
     sig = sig_ap + sig_periodic
 
--- a/neurodsp/sim/combined.py
+++ b/neurodsp/sim/combined.py
@@ -87,53 +87,49 @@ def sim_combined(n_seconds, fs, components, component_variances=1):
 
 
 @normalize
-def sim_central_freq(n_seconds, fs, central_freq, bw, ht, ap_func='sim_powerlaw',
-                     ap_kwargs={'exponent': -2}):
-    """
-    Returns a time series whose full power spectrum consists of an aperiodic component
-    and a gaussian peak at central_freq with standard deviation bw and relative height ht.
-
+def sim_peak_oscillation(sig_ap, fs, freq, bw, height):
+    """Returns a time series whose full power spectrum consists of an aperiodic component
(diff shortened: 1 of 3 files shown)


ryanhammonds:
This is really cool! I fit the psd from your example with fooof to compare chi, central_freq, bw, and ht. Everything matched up as expected, except for a sign flip in the chi (negative here, but fooof reports a positive). Anyways, I'll leave a proper review for this soon.

fooof

TomDonoghue:
This looks awesome Eric - I'm looking forward to trying it out when I get a chance!

The different sign of 1/f is expected based on different conventions between FOOOF & NDSP. Sorta annoying, but sorta grand-fathered in at this point, and consistent with everything else.

elybrand:
@rdgao just wanted to tag you in to let you look at the math if you're so inclined.

rdgao:
holy moly that's a lot of math

ryanhammonds (commented):
>  Do you think I should rewrite this code so that it takes as input a general time series and then adds sinusoids to augment its power spectrum, or do you think that should be a separate function and we can keep this sim as is?

I'm pushed an update so any aperiodic function may be used. However, it defaults to `sim_powerlaw` with an exponent of -2. I also fixed a doc typo, trimmed line lengths for pep8, etc.

neurodsp/sim/combined.py
```diff
@@ -82,3 +84,89 @@ def sim_combined(n_seconds, fs, components, component_variances=1):
...
+    and a gaussian peak at central_freq with standard deviation bw and relative height ht.
+
+
+    Parameters
+    -----------
+    n_seconds: float
+        Number of seconds elapsed in the time series.
+    fs: float
+        Sampling rate.
+    central_freq: float
```
TomDonoghue, line 103: For consistency across periodic functions, I think this can / should just be called `freq`

neurodsp/sim/combined.py
```diff
@@ -82,3 +84,89 @@ def sim_combined(n_seconds, fs, components, component_variances=1):
...
+        Central frequency for the gaussian peak in Hz.
+    bw: float
+        Bandwidth, or standard deviation, of gaussian peak in Hz.
+    ht: float
```
TomDonoghue, line 107: I don't find `ht` a super obvious label. Maybe we just spell it out as `height`.
I'm a little tempted to call it 'power' (maybe 'pow' or even 'pw'), but perhaps that isn't a technically ideal label?
Not totally sure on this, but _maybe_ we want to update this label.

neurodsp/sim/combined.py
```diff
@@ -82,3 +84,89 @@ def sim_combined(n_seconds, fs, components, component_variances=1):
     sig = np.sum(sig_components, axis=0)
 
     return sig
+
+
+@normalize
+def sim_central_freq(n_seconds, fs, central_freq, bw, ht, ap_func='sim_powerlaw',
```
TomDonoghue, line 90: I _think_ maybe we can change the name of this function? I don't find `sim_central_freq` to be particularly obvious (in terms of what this does). Annoyingly, there's not an obvious alternative. 

I think I would be tempted, with a bit of a call-out to FOOOF, to call it something like `sim_peak_oscillation`, or perhaps similar, with the idea that it "simulates an oscillation that exhibits as a peak in the power spectrum, with a designated center frequency, power / height, and bandwidth. 

Thoughts? Any other name suggestions?

neurodsp/sim/combined.py
```diff
@@ -82,3 +84,89 @@ def sim_combined(n_seconds, fs, components, component_variances=1):
...
+        Time series with desired power spectrum.
+
+    Examples
+    --------
+    Simulate an aperiodic component with exponent -2 superimposed over
+    an oscillatory component with central frequency 20.
+
+    >>> sig = sim_central_freq(n_seconds=50, fs=500, central_freq=20, bw=5, ht=7,
+    ...                        ap_func='sim_powerlaw', ap_params={'exponent': -2})
+    """
```
TomDonoghue, line 127: So, this function is something of a special case (as compared to other oscillation funcs), in that it is only available with a sinusoidal basis function. I think this is something perhaps worth mentioning and adding in a `Notes` section, which would say, effectively "note that due to the nature of how this signal is simulated, it is only available with sinusoidal cycles", or similar.

neurodsp/sim/combined.py
```diff
@@ -82,3 +84,89 @@ def sim_combined(n_seconds, fs, components, component_variances=1):
...
+
+    times = create_times(n_seconds, fs)
+
+    # Construct the aperiodic component and compute its Fourier transform. Only use the
+    # first half of the frequencies from the Fourier transform since the signal is real.
+    sig_components = {ap_func: ap_kwargs}
+    sig_ap = sim_combined(n_seconds, fs, sig_components)
```
TomDonoghue, line 134: This doesn't _really_ matter, but I don't think we need to call `sim_combined` here, since we are not actually combining signals. `ap_func` should be a callable, so this should work as:
`sig_ap = ap_func(n_seconds, fs, **ap_kwargs)`

neurodsp/sim/combined.py
```diff
@@ -82,3 +84,89 @@ def sim_combined(n_seconds, fs, components, component_variances=1):
...
+    sig_len = sig_ap.shape[0]
+
+    sig_ap_hat = np.fft.fft(sig_ap)[0:(sig_len//2+1)]
+
+    # Create the range of frequencies that appear in the power spectrum since these
+    # will be the frequencies in the cosines we sum below
+    freqs = np.linspace(0, fs/2, num=sig_len//2 + 1, endpoint=True)
+
+    # Construct the array of relative heights above the aperiodic power spectrum
+    rel_heights = np.array([ ht * np.exp(-(f - central_freq)**2/(2*bw**2)) for f in freqs])
```
TomDonoghue, line 144: style nitpick: I'd use `for freq in freqs` rather than `f` (variables names > len(2), and control-f-able).

neurodsp/sim/combined.py
```diff
@@ -82,3 +84,89 @@ def sim_combined(n_seconds, fs, components, component_variances=1):
...
+                            np.cos(2*np.pi*f*times), 2
+                          )**2 for f in freqs
+                        ])
+
+    # Build an array of the amplitude coefficients
+    cosine_coeffs = np.array([
+                    (-np.real(sig_ap_hat[ell]) + np.sqrt(np.real(sig_ap_hat[ell])**2 + \
+                        (10**rel_heights[ell] - 1)*np.abs(sig_ap_hat[ell])**2))/cosine_norms[ell]
+                    for ell in range(cosine_norms.shape[0])]
+                )
```
TomDonoghue, line 159: holy math :)

I'm not going to pretend I really understand what's going on here. This might be a target for a bit more inline comment description?

also, style comment: multi-line spacing is a bit off (I think - easier to check in text-editor). One of us can run a lint and style check at some point.

neurodsp/sim/combined.py
TomDonoghue: Ahh, okay, so I just noticed that the function is being passed in as a string label, which totally makes sense, but we  can still skip the `sim_combined` step, with this (basically, doing what `sim_combined` does to go from label -> callable:
```
from neurodsp.sim.info import get_sim_func
ap_func = get_sim_func(ap_func) if isinstance(ap_func, str) else ap_func) 
sig_ap = ap_func(n_seconds, fs, **ap_kwargs)
```

Side benefit: this actually generalizes so that `ap_func` can be a string label for a function within the module, or be _any callable_ (so one could pass in custom functions, as long as they return a signal, as expected).

TomDonoghue: Quick comment: add support for array input instead.

ryanhammonds: Just making sure, but replace `ap_func` and `ap_kwargs`, with something like `ap_sig` instead?

ryanhammonds: I think that makes sense. We would need to do additional checks to ensure the `n_seconds` and `fs` are consistent with `ap_sig` to make sure it can be cleanly add to the periodic signal.

neurodsp/sim/combined.py
ryanhammonds: `sim_peak_oscillation` works for me. `sim_gaussian` could also work.

neurodsp/sim/combined.py
ryanhammonds: I have it as `height` for now as I'm working on addressing these comments. `pw` could also work since it matches fooof reports, as does `bw`.

neurodsp/sim/combined.py
ryanhammonds: I added a note.

neurodsp/sim/combined.py
ryanhammonds: Array support added instead.

ryanhammonds:
@TomDonoghue I addressed all of your comments, renamed the func/variables, added array support, ran pylint to resolve multi-line white space issues. I also added an accuracy test to make sure a peak was close to the center freq: `np.argmax(spectrum-spectrum_ap)`.

PR #222, opened 2020-09-03
[ENH] - Simulate time series with knees, sum of sinusoids

elybrand:
## Overview

This simulation builds a time series out of sums of sinusoids whose power spectrum follows the Lorentzian law

P(f) = 1/(f^chi1 * (f^chi2 + k).

Below is a minimal working example.
```python
import numpy as np
from neurodsp.plts import plot_time_series, plot_power_spectra
from neurodsp.sim import sim_knee

n_seconds = 10
fs = 10**3
sig_len=n_seconds*fs
chi1 = 1
chi2= 2
k = 10**2

sig = sim_knee(n_seconds, fs, chi1, chi2, k)
freqs, psd = compute_spectrum(sig, fs)

# Drop the DC component
freqs, psd = freqs[1:], psd[1:]

# Plot against baseline
psd_baseline = np.array([1/(f**chi1*(f**chi2 + k)) for f in freqs])
psd_baseline *= psd[0]/psd_baseline[0]

fig, axes = plt.subplots(2,1, figsize=(15,7))
axes[0].set_title(rf"Knee Signal with Lorentzian $P(f) = \frac{{1}}{{f^{chi1}(f^{chi2} + {k})}}$")
plot_time_series(times, sig, ax=axes[0])
plot_power_spectra(freqs, [psd, psd_baseline], labels=["Empirical PSD", "Lorentzian"], ax=axes[1])
```
knee_signal

## Comments

This code is general enough to allow shaping the power spectrum according to any pre-defined function over frequencies. This could be an extension of the code.

### DIFF
--- a/neurodsp/sim/__init__.py
+++ b/neurodsp/sim/__init__.py
@@ -3,7 +3,7 @@
 from neurodsp.utils.sim import set_random_seed
 
 from .periodic import sim_oscillation, sim_bursty_oscillation
-from .aperiodic import sim_powerlaw, sim_random_walk, sim_synaptic_current, sim_poisson_pop
+from .aperiodic import sim_powerlaw, sim_random_walk, sim_synaptic_current, sim_poisson_pop, sim_knee
 from .cycles import sim_cycle
 from .transients import sim_synaptic_kernel
 from .combined import sim_combined
--- a/neurodsp/sim/aperiodic.py
+++ b/neurodsp/sim/aperiodic.py
@@ -116,6 +116,68 @@ def sim_synaptic_current(n_seconds, fs, n_neurons=1000, firing_rate=2.,
 
     return sig
 
+@normalize
+def sim_knee(n_seconds, fs, chi1, chi2, k):
+    """
+    Returns a time series whose power spectrum follows the Lorentzian equation
+
+    P(f) = 1 / (f**chi1 * (f**chi2 + k))
+
+    using a sum of sinusoids.
+
+    Parameters
+    -----------
+    n_seconds: float
+        Number of seconds elapsed in the time series.
+    fs: float
+        Sampling rate.
+    chi1: float
+        Power law exponent before the knee.
+    chi2: float
+        Power law exponent added to chi1 after the knee.
+    k: float
+        Knee parameter.
+
+    Returns
+    -------
+    sig: 1d array
+        Time series with desired power spectrum.
+
+    Notes
+    -----
+    The slope of the log power spectrum before the knee is -chi1 whereas after the knee it is -(chi1 + chi2).
+
+    Examples
+    --------
+    Simulate a time series with (chi1, chi2, k) = (1, 2, 100)
+
+    >> sim_knee(n_seconds=10, fs=10**3, chi1=1, chi2=2, k=100)
+    """
+
+    times = create_times(n_seconds, fs)
+    sig_len = fs*n_seconds
+
+    # Create the range of frequencies that appear in the power spectrum since these
+    # will be the frequencies in the cosines we sum below
+    freqs = np.linspace(0, fs/2, num=sig_len//2 + 1, endpoint=True)
+
+    # Drop the DC component
+    freqs = freqs[1:]
+
+    # Map the frequencies under the (square root) Lorentzian. This will give us the amplitude coefficients
+    # for the sinusoids.
+    cosine_coeffs = np.array([np.sqrt(1/(f**chi1*(f**chi2 + k))) for f in freqs])
+
+    # Add sinusoids with a random phase shift
+    sig = np.sum(
+                np.array(
+                    [cosine_coeffs[ell]*np.cos(2*np.pi*f*times + 2*np.pi*np.random.rand()) for ell, f in enumerate(freqs)]
+                ),
+                axis=0
+                )
+
+    return sig
+
 
 @normalize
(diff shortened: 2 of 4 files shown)


elybrand:
@rdgao again, just tagging you in case you're curious.

rdgao:
beautiful - thanks @elybrand !

ryanhammonds (commented):
Looks good! I'm pushing an update for pep8 and to fix the failed 3.6 travis build.

TomDonoghue:
This looks awesome, thanks @elybrand! 

My only comments are basically about the API, two open Qs:
- so far, by convention, NDSP uses negative values for the exponent values. I think it would be more consistent with the local module here to converse this convention (otherwise, there is a sign flip when switching out functions within the same module). 
    - it's a bit annoying, since FOOOF has the positive value convention, and this sim is very related to the FOOOF model, but I reckon 'consistency-within-module' > 'consistency-between-modules'
    - I don't think this causes any issues? Just requires some updates to reflect notation. 
- in terms of the input for `chi2`, what do we think about making the user input be the final value for that range (rather than the final value being (ch1 + ch2). This feels more intuitive to me (as in "I want a signal with exponents of 1 & 2", and then  done, rather than, as a user, having to do an extra step of calculating the difference. 
    - I think this wouldn't be a big deal to change. If we do this, can line 171 just use `chi2 - ch1` (or equivalent, depending on sign convention), to do this

TomDonoghue (commented):
in terms of the code, everything looks great to me! Only open things then are comments in main thread.

ryanhammonds:
@TomDonoghue I addressed your first point and flipped the sign of the two chi params in the latest commit.

My understanding for your second point is that we we want a slope of `chi1` before the `knee`(hz) after which point we want a slope of `chi2`. I'm not sure how to accomplish this since all freqs seemed to be affected by `chi1`, `chi2`, and `knee`: 
```
np.array([np.sqrt(1/(f**-chi1*(f**-chi2 + knee))) for f in freqs])
```

Here are some other related issues I've noted:

### Slope before/after the knee

I'll need to write up some code/tests that can check the slope of the spectrum pre/post knee.

Also, I think I'm misunderstanding this note in the docs:
> The slope of the log power spectrum before the knee is chi1 whereas after the knee it is
    (chi1 + chi2).

The example below should produce an exponent of 0 (-2 + 2) after 100hz, but it doesn't:

```
np.random.seed(0)
n_seconds = 10
fs = 1000

sig_knee = sim_knee(n_seconds=n_seconds, fs=fs, chi1=-2, chi2=2, knee=100)
freqs, powers = compute_spectrum(sig_knee, fs)

```
knee

#### Evaluating Knee with FOOOF

I tried to use FOOOF to estimate where the knee occurs but there seems and it's close, but only for some combinations of chi parameters. Below, a signal is simulated with a knee of 100hz and fooof reports 105hz. However if the `chi1` and `chi2` are altered, the `knee` estimated by FOOOF can be very far off:
```
np.random.seed(0)

n_seconds = 10
fs = 1000

# if chi2=-5, the knee estimate by FOOOF becomes 50hz
sig_knee = sim_knee(n_seconds=n_seconds, fs=fs, chi1=-2, chi2=-4, knee=100)
freqs, powers = compute_spectrum(sig_knee, fs)

fm = FOOOF(aperiodic_mode='knee', max_n_peaks=0)
fm.fit(freqs, powers, freq_range=(1, 200))
fm.print_results()
fm.plot(plt_log=True)
```
knee_fooof

elybrand:
@ryanhammonds just to put my two cents in

> It seems all freqs are affected by chi1, chi2, and knee, but my math knowledge is super rusty:
I think what you want to do is just simply redefine what chi2 is. If the user inputs chi2 as the desired slope after the power spectrum, redefining it as
```python
chi2 = chi2-chi1
```
should do the trick. Fundamentally it boils down to how this generalized Lorentzian is behaving asymptotically. Provided the frequency isn't too large, the dominant term in the expression 
```python
 f**chi1 (f**chi2 + k)
```
is `f**chi1`, since `f**chi2 + k ~ k` when `f << k**(1/chi2)`. In the regime where `f > k**(1/chi2)`, the second expression is behaving asymptotically like `f**chi2`, which is why after the knee you expect a slope of `chi1 + chi2`. This whole argument assumes that the signs of both exponents are positive, namely that the power spectrum is *decaying*. When you have exponents with opposite signs, then the slope after the knee is indeed not `chi1 + chi2`. Here, what happens is that `f**chi1 (f**chi2 + k) ~ f**(chi1 + chi2) + kf**(chi1)`. In your example, the term `f**(chi1 + chi2) ~ 1` is no longer the dominant term. Instead, the term `kf**(chi1)` dominates, which is why you're still seeing power law decay.
  
I think this discussion also elucidates why finding the knee when the slope is steep is so hard. Namely, when `chi2 >> 1`,  `f << k**(1/chi2)` is effectively asking that `f < 1`. In other words, there's basically no knee.

ryanhammonds:
@elybrand Thanks for the explanation, it cleared up my confusion. Redefining chi2 as you mentioned should close out this PR then.

Out of curiosity, I wonder how a piecewise function would compare to this implementation. Something like:
```
1/f**chi1 when f <= knee
1/f**chi2 when f > knee
```

elybrand:
@ryanhammonds Yea, I think there are many models which could capture the essence of power spectral knees. I had stuck with this generalized Lorentzian that Richard suggested in Slack because if you set `chi1=0` then it reduces to the model FOOOF uses to fit knees.

Issue #227, opened 2020-11-19
[ENH]: possible new cycle type


TomDonoghue:
This [paper](https://www.frontiersin.org/articles/10.3389/fncom.2016.00087/full) describes a simple type of non-sinuisoidal cycle, which is a bit different from anything we already have available - as it allows for a gaussian like cycle, with a parameter for tweaking the relative amplitude difference of peak / trough. 

The formula, from the paper (formula 9, page 5) is:
`sig = ((cos(2*pi*f*t) + 1) / 2)^b`

Note that everything is pretty standard, except the b parameter, and the plus/divide (which I think just operate to make the data positive, and for scaling). 

This is pretty straight-forward in code:
```
# Set things up
fs = 1000
f = 5
sample = 1500
x = np.arange(sample)

# try new sim
b = 3
vals = ((np.sin(2 * np.pi * f * x / fs) + 1) / 2)**b

# plot sim
plt.plot(x, vals)
```

This gives this outcome:
![Screen Shot 2020-11-19 at 6 44 18 PM](https://user-images.githubusercontent.com/7727566/99737894-415d6480-2a97-11eb-9a3d-23397fa32c82.png)

The parameter `b` is a simple value to change the overall shape of the cycle. 

This is a demonstrated cycle-type in the literature, and is pretty straight-forward, and (I think) extends our functionality, I think we might as well add it. 

@ryanhammonds - when you get a chance, can you:
- have a quick play with this, and compare to our current cycles, especially our existing gaussian cycle, to check / verify that it does add something new and distinct, and then, if so
- prepare an implementation, organizing the formula to match our current organization and API, as we have in `neurodsp/sim/cycles`

TomDonoghue:
I see no reason to not just go for this - but @rdgao might have a more intuitive sense of that formula, and perhaps has quick thoughts and/or wants to give a yay/nay vote?

TomDonoghue:
also, see formula 10, same paper, for a possibly-interesting variant on a sawtooth wave (it's a bit different from what we already have, I think).

rdgao:
these are great! very nice steps towards an analytical description of the different symmetry/shapes.

[...]

PR #228, opened 2020-11-20
[ENH] - Asymmetrical gaussian cycles

ryanhammonds:
I've implemented an asymmetrical gaussian cycle described in #227. The cycles produced from this new function can also be recreated using the `std` param when simulated using `cycle='gaussian'`. It may be more useful to have an optional beta argument for `gaussian` cycles instead? I'm also not sure if this adds something unique (other than a rescaling of `std`).

```
import numpy as np

n_seconds = 1
freq = 5
fs = 500

sig_a = sim_oscillation(n_seconds, fs, freq, cycle="gaussian", std=.1)
sig_b = sim_oscillation(n_seconds, fs, freq, cycle="gaussian_asym", beta=.5, phase=.75)

fig, axes = plt.subplots(nrows=2, ncols=1, figsize=(16, 6))
plot_time_series(times, sig_a, ax=axes[0])
plot_time_series(times, sig_b, ax=axes[1])
```
cyc_1

```
sig_a = sim_oscillation(n_seconds, fs, freq, cycle="gaussian", std=.01)
sig_b = sim_oscillation(n_seconds, fs, freq, cycle="gaussian_asym", beta=15, phase=.75)

fig, axes = plt.subplots(nrows=2, ncols=1, figsize=(16, 6))
plot_time_series(times, sig_a, ax=axes[0])
plot_time_series(times, sig_b, ax=axes[1])
```
cyc_2


TomDonoghue:
Wow, thanks @ryanhammonds for the super quick implementation!

And yeh, that makes sense - I was wondering if it overlapped with our existing Gaussian thing, and it seems it really does!

So I guess, open question - are there cases one can only simulate with one or other, or is this totally redundant? And if they collapse to the same, is there a preferred / more sensible implementation to prioritize?

I think based on Richard's comments in the issue - maybe making the scaling factors into update-able arguments would make the new function allow for more / different kinds of cycles?

I can have a bit of a check with this to get a better sense of if they are totally redundant or not - but also, maybe @rdgao has a better intuition for this?

ryanhammonds:
No problem! I don't think it's completely redundant - I tried to map `std` and `beta` and failed. I went back to Richard's comment and added `shift` and `scale` as kwargs. I also had to adjust the normalize decorator so it wouldn't interfere. The last thing to do is add is tests.

Here is are plots varying the `shift` and `scale` parameters:
shift_scale

ryanhammonds:
@TomDonoghue Sorry I took so long the get back to this! I tried to implement equation 10 in the last commit. However, there is an issue with this implementation and I haven't figure out a fix. The signal when fs > 1 cycles are cutoff:

```

n_seconds = 5
fs = 500

phi = 3*np.pi/2

freq = [1, 2]
sig_a = sim_oscillation(n_seconds, fs, freq[0], cycle="harmonic_asym", phi=phi, n_harmonics=2)
sig_b = sim_oscillation(n_seconds, fs, freq[1], cycle="harmonic_asym", phi=phi, n_harmonics=2)

fig, axes = plt.subplots(nrows=2, ncols=1, figsize=(18, 6))
plot_time_series(times, sig_a, ax=axes[0], title="freq=1")
plot_time_series(times, sig_b, ax=axes[1], title="freq=2")
```
plots

neurodsp/sim/cycles.py
```diff
@@ -204,6 +204,109 @@ def sim_gaussian_cycle(n_seconds, fs, std):
     return cycle
 
 
+def sim_gaussian_asym_cycle(n_seconds, fs, beta, scale=2, shift=1):
```
TomDonoghue, line 207: we need to change the name on this, I think. It ends 'gaussian-like', but that's not how it's generated. I'm not sure what the best name is though...

neurodsp/sim/cycles.py
```diff
@@ -204,6 +204,109 @@ def sim_gaussian_cycle(n_seconds, fs, std):
...
+        Translate the signal along the y-axis.
+
+    Returns
+    -------
+    cycle : 1d array
+        Simulated asymmetrical gaussian cycle.
+
+    Notes
+    -----
+    If shift or scale keyword arguments are given, default signal normalization will be bypassed.
```
TomDonoghue, line 236: this note seems somewhat misplaced, because using this function directly, none of our cycles are normalized. This only applies to simulating this cycle with `sim_oscillation`. We probably want to note that _somewhere_, but I don't think here.

neurodsp/sim/cycles.py
```diff
@@ -204,6 +204,109 @@ def sim_gaussian_cycle(n_seconds, fs, std):
...
+    Examples
+    --------
+    Simulate an asymmetrical cycle as the sum of harmonics:
+
+    >>> cycle = sim_gaussian_asym_cycle(n_seconds=1, fs=500, beta=3)
+    """
+    times = create_cycle_time(n_seconds, fs)
+    cycs = np.zeros((n_harmonics+1, len(times)))
+
+    harmonics = np.array([harmonic for harmonic in range(1, n_harmonics + 2)])
```
TomDonoghue, line 300: I think this can be simplified to `np.array([range(1, n_harmonics + 2)])`?

neurodsp/sim/cycles.py
```diff
@@ -204,6 +204,109 @@ def sim_gaussian_cycle(n_seconds, fs, std):
...
+    ----------
+    Lozano-Soldevilla, D., Huurne, N. T., &amp; Oostenveld, R. (2016). Neuronal Oscillations with
+    Non-sinusoidal Morphology Produce Spurious Phase-to-Amplitude Coupling and Directionality.
+    Frontiers in Computational Neuroscience, 10. doi:10.3389/fncom.2016.00087
+
+    Examples
+    --------
+    Simulate an asymmetrical cycle as the sum of harmonics:
+
+    >>> cycle = sim_gaussian_asym_cycle(n_seconds=1, fs=500, beta=3)
```
TomDonoghue, line 295: this example is for the wrong cycle

TomDonoghue:
Okay, I had a review on this, leaving some quick comments. 

Everything looks good - my main thoughts are 1) naming, and 2) potential extensions (like I'm thinking about maybe updating the second approach to take a specified amplitude for each simulated wave). I think I'll have to play with these cycles a bit for that though. 

@ryanhammonds - can you check the comments, update, and see if there is anything else you think should be done here, and then I'll come back and play with these a bit. 

Also: ToDo - add the new functions to `doc/api.rst`

neurodsp/sim/cycles.py
ryanhammonds: Maybe ``sim_exp_asym_cycle`` or `sim_beta_asym_cycle`` since it's being raised to beta?

neurodsp/sim/cycles.py
ryanhammonds: I can move it to `sim_oscillation` and `sim_bursty_oscillation`?

neurodsp/sim/cycles.py
ryanhammonds: good catch, thanks!

neurodsp/sim/cycles.py
ryanhammonds: Good idea! I didn't realize ranges could be passed like this. One note for when I come back to make this round of updates: the brackets will need to be removed so it's not a 2D array.

neurodsp/sim/cycles.py
ryanhammonds: Updated to `sim_asym_beta_cycle` and `sim_asym_harmonic_cycle`. Feel free to change.

neurodsp/sim/cycles.py
ryanhammonds: I moved this to a note in the normalize decorator.

ryanhammonds:
@TomDonoghue, I came back around to this and pushed updates related to your review. I also merge main / fixed a mess of conflicts.

I think everything should almost be ready. The one thing that's left is to maybe update the func names and add to the api.

neurodsp/sim/cycles.py
```diff
@@ -300,6 +300,112 @@ def sim_skewed_gaussian_cycle(n_seconds, fs, center, std, alpha, height=1):
...
+def sim_asym_beta_cycle(n_seconds, fs, beta, scale=2, shift=1):
+    """Simulate an asymmetrical guassian-like cycle.
+
+    Parameters
+    ----------
+    n_seconds : float
+        Length of cycle window in seconds.
+    fs : float
+        Sampling frequency of the cycle simulation.
+    beta : float
```
TomDonoghue, line 312: suggestion: rename beta -> exp (avoid name of freq band).

neurodsp/sim/cycles.py
```diff
@@ -300,6 +300,112 @@ def sim_skewed_gaussian_cycle(n_seconds, fs, center, std, alpha, height=1):
     return cycle
 
 
+def sim_asym_beta_cycle(n_seconds, fs, beta, scale=2, shift=1):
```
TomDonoghue, line 303: suggestion: `sim_exp_cos_cycle`

PR #230, opened 2020-12-30
[ENH] - Update rhythm code & docs

TomDonoghue:
This PR is for updates to the `rhythm` module, including code documentation, and tutorials. 

So far, it updates the tutorials (since they were a bit limited) and in code documentation. 

During this PR, I noticed an issue I noticed in SWM, that was also mentioned by Eric in #193 about how it does the window spacing. I think we need to audit and update this code a bit. 
@ryanhammonds : could you do a check through the SWM code algorithm (see paper here: https://github.com/bartgips/SWM, and Matlab implementation here: https://github.com/bartgips/SWM), and see what you think might need fixing? There are a couple notes in https://github.com/neurodsp-tools/neurodsp/issues/193 about it as well. 
From my check, I think at the window check should ensure samples are less than X samples apart from another (not greater than). I think there should perhaps also be an additional check to ensure windows are distinct time segments. 

As a sidenote, I'm not really sure that SWM needs to return costs, so maybe we can update to make that an optional return, and move toward deprecating that output.


ryanhammonds:
I dug into this also think that there is an issue with the current implementation. Eric's comment in the #193 ("You need to select from indices which are at least `spacing_n_samps` away from the entire window to ensure they are disjoint.") _maybe_ incorrect based on Algorithm 1 of the [manuscript](https://github.com/bartgips/SWM/blob/master/Manuscript.pdf):

swm_2

Based on the "initialization" section above, `spacing_n_samps` (or `G` in the manuscript) appears independent of the window length, `L`.

In Eric's code example in #193, the `window_start` indices:
```
len_sig = 23
window_length = 10
n_samp = len_sig - window_length
window_starts = np.array([0, 13])
spacing_n_samps = 3
```
should be "separated by twice the minimal spacing", so I think the indices should be:
```
window_starts = array([0, 6, 12, 18])
```
Then, for any position (Xn), a new/random position (Xn') must be at least `G` (or window_length) away from Xn-1 and Xn+1, based on Algorithm 1 and Fig 2 from the manuscript:

swm

So I _think_ this is what the new window func should be.
```
current_window_idx = window_starts_temp[window_idx_replace]

def _find_new_window_idx(window_starts, spacing_n_samps, current_window_idx, tries_limit=1000):
    """Find a new sample for the starting window."""

    for n_try in range(tries_limit):

        # Generate a random sample & check how close it is to other window starts
        new_samp = np.random.randint(current_window_idx-spacing_n_samps,
                                     current_window_idx+spacing_n_samps)
        
        # Find adjacent window starts
        prev_idx = current_window_idx - (2*spacing_n_samps)
        next_idx = current_window_idx + (2*spacing_n_samps)
        
        # Ensure the new, random sample is at least spacing_n_samps 
        #   away from the next and following window starts.
        dists = np.array([np.abs(window_starts[prev_idx] - new_samp),
                          np.abs(window_starts[next_idx] - new_samp)])

        # If the random sample if too close to an adjacent window start,
        #   raise an error.
        if np.max(dists) < spacing_n_samps:
            break

    else:
        raise RuntimeError('SWM algorithm has difficulty finding a new window. \
                            Try increasing the spacing parameter.')

    return new_samp
```

I'm going to have to revisit this because I could be completely wrong. I took a peek at the matlab implementation and it's going to take me a bit to parse through.

ryanhammonds:
@TomDonoghue  I pushed an algorithm update in 54592dc.

[Here](https://github.com/ryanhammonds/neurodsp/blob/swm/SWM.ipynb) is a notebook I used for testing the update and comparing the results to the matlab implementation. It seems to recover average window better than the matlab version. Before the update, the example in the notebook would return an error.

The version I pushed really only works for a small number of windows (i.e. short signals / large window length and spacing) and a small number of iterations. The computation time exponentially increases with the number of windows. I think this is because it brute forces distances (i.e. cost difference) between every combination of windows, rather than using Markov Chain Monte Carlo sampling like in the paper / matlab version.

If we wanted to use MCMC sampling, I think we should add an external dependency unless we wanted to add at `stats` module to ndsp. I tried bootstraping + multiprocessing when computing the distance between pairs to speed things up couldn't get it to work. I can try again though.

neurodsp/rhythm/swm.py
```diff
@@ -149,17 +170,46 @@ def _compute_cost(sig, window_starts, win_n_samps):
...
+    Returns
+    -------
+    new_samp : int
+        The index for the new window.
+    """
+
+    for _ in range(tries_limit):
+
+        # Find adjacent window starts
+        prev_idx = current_window_idx - (2*spacing_n_samps)
```
TomDonoghue, line 196: my main concern is that this assumes the window positions based on their original positions, right? Which works in the first iteration, but as multiple windows have been moved, this seems like it would no longer guarantee non-overlap, right? Like, if the previous window has moved, checking the overlap based on it's original position might not be enough. 

So, I think this function should actually take in the full array of current window positions, and compare to that?

neurodsp/rhythm/swm.py
```diff
@@ -19,20 +19,20 @@ def sliding_window_matching(sig, fs, win_len, win_spacing, max_iterations=500,
...
     win_spacing : float
-        Minimum window spacing, in seconds.
-    max_iterations : int
+        Minimum spacing between windows, in seconds. This is G in the original paper.
+    max_iterations : int, optional, default: 500
         Maximum number of iterations of potential changes in window placement.
-    temperature : float
-        Temperature parameter. Controls probability of accepting a new window.
+    temperature : float, optional, default: 1
+        Controls the probability of accepting a new window.
```
TomDonoghue, line 28: This is a totally optional extension - if you want: but I do find it a little weird that "win_spacing" generally determines the number of windows. In the original, there is an option to set a specific number of windows. 
What do you think about extending `window_starts_custom` to not only take an array (as current), but also take an int, and if so, that is the the number of windows that are used (that get initialized as evenly spaced). I think this would be easy to add?

TomDonoghue:
Thanks for digging into this one @ryanhammonds! 

In terms of comparing our version to the Matlab version - we have a minimal version. I don't think it's worth it right now to massively extend this algorithm (unless / until someone really needs it) - so I vote not trying to add MCMC etc at the moment. To be more clear about our implementation, I've added some description to the docstring, and fleshed out sub-function descriptions to try and better track what's going on. 

Being clear this is a minimal implementation, I think as long as we are confident about what it does, we should merge this as is. I've left a couple review comments on the code for things I'm not clear on - once those are done, then this should be good to go without major extension.

neurodsp/rhythm/swm.py
ryanhammonds: As new positions are accepted, the new positions are updated before the next iteration. Lines 114-118 above:
```
if np.random.rand() < p_accept:

    # Update costs & windows
    costs[iter_num] = cost_temp
    window_starts = window_starts_temp
```

TomDonoghue: yeh - but that's in the main function. The `_find_new_window_idx` doesn't use the `window_starts` variable, it only takes the current window, then infers the location of adjacent windows (lines 196-197), right? Within the context of `_find_new_window_idx`, it doesn't to have info about where the other windows actually are, I think? Am I completely missing something?

ryanhammonds: Oh I see now! Yeah, I think we want what you described. I'll update and see how if the results change.

ryanhammonds: Okay, so I updated `_find_new_window` to use actual window locations instead of inferring. However, the implementation can't find pure gaussians with random amounts of white noise between bursts (here's a [notebook](https://github.com/ryanhammonds/neurodsp/blob/swm/SWM.ipynb)). I'm lost as to why this is the case - it's probably a bug or a misunderstanding of the algorithm. 

Pre  367ebc2:
pre

Post 367ebc2:
post

Maybe we should have a notebook with a few ideal cases to ensure the implementation here can do as well (or better) than the matlab version.

How SWM converges to a pattern is a bit confusing at it feels like a brute force hack that requires lots of computational power even with fancy (MCMC) resampling techniques. Even the matlab swm motif found in the notebook above is a a very messy pattern considering pure oscillations are simulated. Maybe we the new napp + bycycle approach could be a simpler alternative :wink:

neurodsp/rhythm/swm.py
ryanhammonds: This would basically be the same as using:
```
win_spacing=len(sig)/n_spaces
```
where `n_spaces` it the int you wanted to add as an optional param), right?

TomDonoghue: Yeh, instead of overloading `window_starts_custom`, an optional parameter of `n_spaces` could be added (though I think I'd say this could / should be called `n_windows`). In the case of passing in this parameter, then `win_spacing` wouldn't have to be explicitly set, and could be computed as you have above. 

At least, that's how I remember it from the original implementation. The Matlab code might have a different API for setting these things (which I don't think we need to explicitly follow) - but the option to set the number of windows is something that is described in # 2 of the algorithm description in the paper.

neurodsp/rhythm/swm.py
TomDonoghue: Hmm - that's a bit weird. In the simulated case, the "events" are right next to each other, so the algorithm should theoretically do better with a reduced value for `win_spacing` - does that help by any chance. 

Broadly - I don't want to get too pulled into a full audit of the method here. I think we should do diligence that (to the best of our knowledge) the algorithm proceeds as described in the paper, and is not obviously way worse than the matlab version. I think that's is broadly the case? (At least, the Matlab version in that notebook also doesn't do great on the noisy gaussian case). If you have a few minutes can you do a couple comparisons to matlab? If it looks comparable, I think we stick with this. 

I agree that how SWM converges is fairly brute force, and that TBH, it's probably not the best way to go - I would be somewhat more optimistic about spectral + temporal parameterization :)

neurodsp/rhythm/swm.py
```diff
@@ -44,14 +44,19 @@ def sliding_window_matching(sig, fs, win_len, win_spacing, max_iterations=500,
...
-    - Parameters `win_len` and `win_spacing` should be chosen to be about the size of the
-      motif of interest, and the N derived should be about the number of occurrences.
+    - The `win_len` parameter should be chosen to be about the size of the motif of interest.
+      The larger this window size, the more likely the pattern to reflect slower patterns.
+    - The `win_spacing` parameter also determines the number of windows that are used.
+    - If looking at high frequency activity, you may want to apply a highpass filter,
+      so that the algorithm does not converge on a low frequency motif.
+    - This implementation is a minimal version, as compared to the original implementation
+      in [2], which has more available options.
+    - This version may also be much smaller than the original, as this implementation does not
```
TomDonoghue, line 58: I _think_ this should say "slower" rather than "smaller"?

neurodsp/rhythm/swm.py
ryanhammonds: I missed your response here, sorry. I've ran the matlab version and the results are considerably better than after the algo update in 367ebc2. I've re-ran this code and what is currently here produces unstable results. Sometimes the waveform is a noisy gaussian and other times it's white noise. Reducing `win_spacing` didn't improve results either.

I'll have to dig into this deeper to see if there are any obvious bugs. The algorithm described in the paper seems relatively simple, but yet the matlab code is 2000+ lines long, and does a lot more than what is described in the algorithm section of the paper, so maybe I'm missing some other key part to the implementation.

ryanhammonds:
@TomDonoghue, I rebased main and resolved conflicts here. I also re-wrote the swm implementation with several modifications from the original to get things working. The main deviations from the published swm algorithm are:

1. Each iteration is similar to an neural net epoch, randomly moving _all_ windows, in random order. The original implementation does not guarantee even window sampling.
2. New window acceptance is determined via increased correlation coefficients and reduced variance across windows. Before, this was a cost measure that didn't work.
3. Phase optimization / realignment to escape local minima.
4. Remove low variance windows to speed up runtime.

These changes allow the algorithm to scale to much longer timeseries.

TomDonoghue (commented):
Looks good!

ryanhammonds:
Thanks for helping clean the rebase up! As for the questions:

- The 2d array is intended. I think it's a bit more flexible this way - for things like the std/var across windows or for clustering windows.
- I pushed an update to the tutorial. The double peak was basically just noise and had very low variance. What was happening was that the spacing param was large relative to the window and the default number of iterations (100) wasn't enough to search the wide space between windows to find a proper motif. I ended up just lowering the the spacing to equal the window length to address this, and added a var_thresh to speed the search up.

TomDonoghue:
@ryanhammonds - this all makes sense to me! I re-ran the tutorial, and it seems to pull out a reasonable peak. 

This PR should be ready to merge, I think!

PR #231, opened 2020-12-31
[DOC] - Add tutorials for aperiodic sub-module

TomDonoghue:
This is a placeholder / WIP PR for adding tutorials for the aperiodic module. 

@ryanhammonds : can you have a first go at developing these tutorials? They can start as pretty simple, showing code application, and then I can check and fill in any details later.


ryanhammonds:
Hi @TomDonoghue, these tutorials should be ready for review. 

- DFA: Compares DFA exponent between white noise and power law signals.
- IRASA: Runs algorithm on a 10hz oscillation + -2 exp powerlaw.
  - I added a relative threshold to IRASA that allows cleaner separation of aperiodic/periodic components in some cases. This may need to be reworked or can be removed if not appropriate. I thought it was useful but there may be implications I haven't considered.
- Autocorr: Computes autocorrs of a sine wave and white noise.
  - I updated the module func to include `max_lag`. Before, if `max_lag == 1000`, the func would compute up to the 999 lag. I've changed this so the func is inclusive of `max_lag`. If this isn't what we want, I can change it back.

TomDonoghue:
@ryanhammonds - awesome, thanks for your work on this one!

In terms of code updates, the autocorr lag length is a good catch! The threshold for IRASA is interesting, I'm not sure if it might have some quirks in some cases, but given it defaults to off, and helps in some cases, then I think it's a cool / useful things to keep. 

For the tutorials, I went through and did direct edits, and added a bit of stuff. The only major change is the autocorr, in that I wasn't really clear where the phase comparisons were going, so I switched that up (if I seem to have missed something there, let me know, and we can perhaps add some stuff back). 

When you get a chance, can you review the updates to the tutorials? If they look good to you, then I think this one is done!

tutorials/aperiodic/plot_Fluctuations.py
```diff
@@ -0,0 +1,153 @@
...
+# Import the function for computing fluctuation analyses
+from neurodsp.aperiodic import compute_fluctuations
+
+###################################################################################################
+# Detrended Fluctuation Analysis
+# ------------------------------
+#
+# Detrended fluctuation analysis is a method for analyzing the self-similarity of a signal.
+#
+# It is in some ways similar to looking at autocorrelation, and can is typically used to look
```
ryanhammonds, line 32: typo

tutorials/aperiodic/plot_Fluctuations.py
```diff
@@ -0,0 +1,153 @@
...
+# - :math:`\alpha = 1.0` : 1/f
+# - :math:`1 < \alpha < 2` : correlated (non-stationary)
+#
+
+###################################################################################################
+# Applying DFA
+# ------------
+#
+# Here, to introduce DFA, we will use colored noise signals, (white noise and pink noise),
+# with have different auto-correlation properties, and so should show different DFA results,
```
ryanhammonds, line 60: typos

tutorials/aperiodic/plot_Fluctuations.py
```diff
@@ -0,0 +1,153 @@
...
+# which we can compare.
+#
+# Note that DFA can be applied to multiple signal types, and in neural field data it is
+# often used to examine amplitude time series of neural oscillations.
+#
+
+###################################################################################################
+
+# Simulation settings settings
```
ryanhammonds, line 69: first bursts bursts and now settings settings :wink:

tutorials/aperiodic/plot_Fluctuations.py
```diff
@@ -0,0 +1,153 @@
...
+###################################################################################################
+
+_, ax = plt.subplots(figsize=(5, 5))
+
+ax.loglog(ts_wn, flucs_wn, label="White Noise")
+ax.loglog(ts_pl, flucs_pl, label="Power Law")
+
+ax.set(title="Fluctuations Across Window Sizes",
+       xlabel="Time Scales", ylabel="Fluctuations")
+plt.legend();
```
ryanhammonds, line 135: Why is a semi-colon here?

tutorials/aperiodic/plot_IRASA.py
```diff
@@ -0,0 +1,161 @@
...
+n_seconds = 10
+fs = 500
+
+# Define the parameters of the simulated components
+cf = 10
+exp = -2
+
+# Define the components for the simulated signal
+components = {'sim_oscillation' : {'freq' : cf},
+			  'sim_powerlaw' : {'exponent' : exp}}
```
ryanhammonds, line 60: whitespace

tutorials/aperiodic/plot_IRASA.py
```diff
@@ -0,0 +1,161 @@
...
+#
+
+###################################################################################################
+# Decomposition
+# -------------
+#
+# Note that what IRASA returns is a decomposition of the power spectrum, separating
+# aperiodic and periodic components.
+#
+# To verify that this is what the the algorithm does, we can check that the spectrum
```
ryanhammonds, line 132: the the

ryanhammonds (commented):
Hi @TomDonoghue, I found a few typos but other than that it looks good.

The phase comparison tried to show how the phase changes as a result of lagging the signal. The second removed plot was representative of the correlation between lags at difference points on the first plot. No worries about removing it though. If it wasn't clear it shouldn't go in.

tutorials/aperiodic/plot_Fluctuations.py
TomDonoghue: fixed

tutorials/aperiodic/plot_Fluctuations.py
TomDonoghue: fixed.

tutorials/aperiodic/plot_Fluctuations.py
TomDonoghue: oh no. word are hard :(

tutorials/aperiodic/plot_Fluctuations.py
TomDonoghue: these mpl utility functions return some nonsense, which gets printed out in the sphinx site and/or downloaded notebooks. This is just to supress output.

tutorials/aperiodic/plot_IRASA.py
TomDonoghue: i think there's an echo in my head.

TomDonoghue:
Thanks for the review @ryanhammonds! I've updated the files, and I think this one is good to go now. 

Also: ahhh, I think I've just tracked the phase delay thing of the autocorr example - showing different delays of the same signal. I think that's an interesting demo, that would perhaps just need a bit more explanation. If you ever feel like integrating it back in, feel free to update & PR!

Issue #229, opened 2020-12-30
[DOC] - Split up & extend filter tutorial(s)


TomDonoghue:
I think it would be a good idea to re-organize and expand our Filter tutorial, which is currently one tutorial page.

A potential reorganization would split up into: FIR, IIR, and maybe an additional tutorial on the utils / plots available to explore filters. 

There are some good resources on guiding the use of filters, that include demos / ideas that we could add in code:
- [de Cheveigne et al, 2019: Filters: when, why and how](https://doi.org/10.1016/j.neuron.2019.02.039)
- [Widmann et al 2015: practical digital filter design](https://doi.org/10.1016/j.jneumeth.2014.08.002)

The above linked papers have a bunch of specific and practical recommendations. With links to these papers as source, I think we could flesh out our discussion of filters to recommend these points, and demonstrate some of the issues they discuss. 

Somewhat related is the idea of a Filter report (#139). Even if this isn't an explicit / separate thing, the tutorials should cover all information that should be reported by a filter, including: filter type, frequency parameters (cutoff frequency), filter order (or length), roll-off or transition bandwidth, passband ripple & stopband attenuation, filter delay (zero-phase, linear phase, etc), causality, and whether it was applied in one or two directions. 

Widmann et al compare some software implementations, and also have suggestions for the user (things we should make sure are easy to do). One in particular: examine filter responses by applying them to impulses (compute the impulse response). This is something we can emphasize in the tutorials, consistently showing and doing this.

TomDonoghue:
This topic is covered in #232, which is being finished to merge, so this issue can be closed now.

PR #232, opened 2020-12-31
[DOC] - Update filter tutorial(s)

TomDonoghue:
This is a WIP start to updating the filter documentation / tutorials, relating to #229. 

@ryanhammonds : following the issue, can you have a go working on this PR? I would suggest digging into the linked papers to guide some extra detail that can be added. If you can organize the new layout, add some detail and extend and you think is needed, I can come back to it after and edit. Thanks!


ryanhammonds:
Hi @TomDonoghue, I still have more work to do on this PR, but I made a first attempt at an FIR tutorial. The Widmann 2015 recommendations are in included in the tutorial, so #235 has been squashed into 48a688c.

TomDonoghue:
I had a quick look at this one. It seems that the math isn't rendering properly? Also, I'm not sure it's particularly needed to get into the details of the math on this. We can't fully cover the technicalities of filters here, and I think more code examples would be better. Also, more general description, introducing, for example what an 'FIR' filter, in particular, is, and the different passtypes (though that might be in a more a different tutorial). 

Also, in terms of organizing the tutorials, I'm thinking maybe:
- 0: filtering in general (introduce passtypes)
- 1: FIR (introduce specifics of FIR filters)
- 2: IIR (introduce specifics of IIR filters)

ryanhammonds:
I'll remove the math - it helped me understand FIR filters and thought it may be useful to others. I agree with the organization. That was the plan I had in mind too. I'm still working on the IIR tutorial but it'll be finished soon.

TomDonoghue:
@ryanhammonds - if you found the math to be helpful to explain what's up with filters, then maybe we should keep it! I found it hard to get a sense of when it wasn't rendering - but when it prints properly, it might be useful, so feel free to keep it for now!

ryanhammonds:
Sounds good! The math is fixed by changing `..math` to `.. math`. I have this locally but will push to soon.

TomDonoghue:
Okay - I finally came back around to this one. The stuff you added and organized looks great @ryanhammonds, and I also came around to keeping the math! I kept pushing on it a bit, with the main theme of organizing what's where, and avoiding duplication - so the I split out more stuff from the original tutorial, added a fourth section for "filter checks", and edited from there, trying to edit in more information as I could. Thanks for doing a push on this, it really helped to have a big chunk of this done. 

This should be ready to merge, I think. @ryanhammonds - when you get a chance, if you could do a quick scan for any typos / bugs / or word quirks, that would be great. If it looks good, we can merge.

ryanhammonds (approved):
This looks good! I'm pushing commit to correct a few typos I found. One thing we may want to do later is add a filter report section to the end of the "checks" tutorial once #235 is merged.

TomDonoghue:
Awesome, thanks for the check @ryanhammonds - the fixes look great!

And yeh, we should definitely update this to reflect the Filter Report once #235 is ready, I'm going to try to get to that soon. 

Merging this in now!

PR #233, opened 2021-01-02
fixing bug where plot_power_spectra wasnt using color provided

meredithjanemccarty:


TomDonoghue:
Hey @meredithjanemccarty - great catch on this, and thanks for the fix!

Right now the tests are failing, and I think it's just because in the case that `colors=None` the zip iteration fails.

In this line, just above what was changed:
```
if colors is not None:
    colors = repeat(colors) if not isinstance(colors, list) else cycle(colors)
```
we don't actually deal with the scenario of colors being None, which was our mistake. Previously, it didn't really matter, because we weren't actually using the `colors` variable (oops). 

I think all that is needed to make this work is to remove the `if colors is not None:` conditional. That way, if colors is None, then it will apply the `repeat`, and all shall be good. 

Do you want to make this quick update? It would just be another commit, removing line 58, unindenting 59, and then checking if it all works. If you prefer, I can handle this update. 

Either way, as long as the tests pass after, this looks good to merge in! Thanks!

meredithjanemccarty:
Went ahead and removed the conditional. Tested it locally, and all works fine. Do you have any documentation on running your test suite locally for any further changes? 

Thank you for all your work making this module open source!

ryanhammonds:
Hi @meredithjanemccarty, thanks for the PR. As for testings, I recommend:

```
pip install pytest pytest-cov
pytest --cov=neurodsp --cov-report=term-missing neurodsp 
```

TomDonoghue:
Thanks a lot for the fix @meredithjanemccarty - this all looks good, and I'm going to merge it in now!

@ryanhammonds - we do have a contrib doc (https://github.com/neurodsp-tools/neurodsp/blob/master/CONTRIBUTING.md), but I just checked and we don't suggest a best way to run the tests locally - so updating to add that would be great whenever you have a chance!

ryanhammonds:
@TomDonoghue Oh whoops, totally forgot about that. Thanks!

I'll add a testing section next week. 

Hope you had a good new years! 😀

Issue #139, opened 2019-02-25
Filter Report


TomDonoghue:
Throwing something here from a recent discussion with @voytek (maybe you want to flesh out this idea / add what you were thinking if it's more / different than I remember). 

It would be nice if, when applying filter, you could generate a quick report about the properties of the filter you applied that would, for example, plot the frequency response, pass & transition band properties and some meta-data. This could be printed out for checking, and optionally saved, to be checked later. 

This could be broadly similar to how FOOOF generates reports for model fitting. With the refactor of filter code (#132), everything should be very easy to use to do this - it's just an extra function to use these functions to generate the report, if requested.

PR #235, opened 2021-01-13
[WIP/ENH] Filter Report

ryanhammonds:
Related to #229 and #139. This PR adds a `verbose` arg to `filter_signal`, `filter_signal_fir` and `filter_signal_iir` that prints out the filter information (suggested in Widmann et al 2015). This includes:

- Filter type (high-pass, low-pass, band-pass, band-stop, FIR, IIR)
- Cutoff frequency (including definition)
- Filter order (or length)
- Roll-off or transition bandwidth
- Passband ripple and stopband attenuation
- Filter delay (zero-phase, linear-phase, non-linear phase) and causality
- Direction of computation (one-pass forward/reverse, or two-pass forward and reverse)

The previous `print_transitions` argument functions the same as the verbose arg now (i.e. `verbose=np.any([print_transitions, verbose])`). I can fully deprecate this arg or modify it to only print transitions.

I'm not sure if the last three bullets are determined correctly (they probably aren't). I'll have to revisit to sort this out.

### DIFF
--- a/neurodsp/filt/checks.py
+++ b/neurodsp/filt/checks.py
@@ -90,7 +90,7 @@ def check_filter_definition(pass_type, f_range):
 
 
 def check_filter_properties(filter_coefs, a_vals, fs, pass_type, f_range,
-                            transitions=(-20, -3), verbose=True):
+                            transitions=(-20, -3), filt_type=None, verbose=True):
     """Check a filters properties, including pass band and transition band.
 
     Parameters
@@ -164,7 +164,8 @@ def check_filter_properties(filter_coefs, a_vals, fs, pass_type, f_range,
 
     # Compute pass & transition bandwidth
     pass_bw = compute_pass_band(fs, pass_type, f_range)
-    transition_bw = compute_transition_band(f_db, db, transitions[0], transitions[1])
+    transition_bw, f_range_trans = compute_transition_band(f_db, db, transitions[0], transitions[1],
+                                                           return_freqs=True)
 
     # Raise warning if transition bandwidth is too high
     if transition_bw > pass_bw:
@@ -174,8 +175,50 @@ def check_filter_properties(filter_coefs, a_vals, fs, pass_type, f_range,
 
     # Print out transition bandwidth and pass bandwidth to the user
     if verbose:
-        print('Transition bandwidth is {:.1f} Hz.'.format(transition_bw))
-        print('Pass/stop bandwidth is {:.1f} Hz.'.format(pass_bw))
+
+        # Filter type (high-pass, low-pass, band-pass, band-stop, FIR, IIR)
+        print('Pass Type: {pass_type}'.format(pass_type=pass_type))
+
+        # Cutoff frequency (including definition)
+        cutoff = round(np.min(f_range) + (0.5 * transition_bw), 3)
+        print('Cutoff (half-amplitude): {cutoff} Hz'.format(cutoff=cutoff))
+
+        # Filter order (or length)
+        print('Filter order: {order}'.format(order=len(f_db)-1))
+
+        # Roll-off or transition bandwidth
+        print('Transition bandwidth: {:.1f} Hz'.format(transition_bw))
+        print('Pass/stop bandwidth: {:.1f} Hz'.format(pass_bw))
+
+        # Passband ripple and stopband attenuation
+        pb_ripple = np.max(db[:np.where(f_db < f_range_trans[0])[0][-1]])
+        sb_atten = np.max(db[np.where(f_db > f_range_trans[1])[0][0]:])
+        print('Passband Ripple: {pb_ripple} db'.format(pb_ripple=pb_ripple))
+        print('Stopband Attenuation: {sb_atten} db'.format(sb_atten=sb_atten))
+
+        # Filter delay (zero-phase, linear-phase, non-linear phase)
+        if filt_type == 'FIR' and pass_type in ['bandstop', 'lowpass']:
+            filt_class = 'linear-phase'
+        elif filt_type == 'FIR' and pass_type in ['bandpass', 'highpass']:
+            filt_class = 'zero-phase'
+        elif filt_type == 'IIR':
+            filt_class = 'non-linear-phase'
+        else:
+            filt_class = None
+
+        if filt_type is not None:
+            print('Filter Class: {filt_class}'.format(filt_class=filt_class))
+
+        if filt_class == 'linear-phase':
+            print('Group Delay: {delay}s'.format(delay=(len(f_db)-1) / 2 * fs))
+        elif filt_class == 'zero-phase':
+            print('Group Delay: 0s')
+
+        # Direction of computation (one-pass forward/reverse, or two-pass forward and reverse)
+        if filt_type == 'FIR':
+            print('Direction: one-pass reverse.')
+        else:
+            print('Direction: two-pass forward and reverse')
 
     return passes
 
--- a/neurodsp/filt/fir.py
+++ b/neurodsp/filt/fir.py
@@ -14,7 +14,8 @@ from neurodsp.filt.checks import (check_filter_definition, check_filter_properti
 ###################################################################################################
 
 def filter_signal_fir(sig, fs, pass_type, f_range, n_cycles=3, n_seconds=None, remove_edges=True,
-                      print_transitions=False, plot_properties=False, return_filter=False):
+                      print_transitions=False, plot_properties=False, return_filter=False,
+                      verbose=False):
     """Apply an FIR filter to a signal.
 
     Parameters
@@ -48,6 +49,8 @@ def filter_signal_fir(sig, fs, pass_type, f_range, n_cycles=3, n_seconds=None, r
         If True, plot the properties of the filter, including frequency response and/or kernel.
     return_filter : bool, optional, default: False
         If True, return the filter coefficients of the FIR filter.
+    verbose : bool, optional, default: False
+        If True, print out detailed filter information.
 
     Returns
     -------
@@ -78,7 +81,8 @@ def filter_signal_fir(sig, fs, pass_type, f_range, n_cycles=3, n_seconds=None, r
     check_filter_length(sig.shape[-1], len(filter_coefs))
 
     # Check filter properties: compute transition bandwidth & run checks
-    check_filter_properties(filter_coefs, 1, fs, pass_type, f_range, verbose=print_transitions)
+    check_filter_properties(filter_coefs, 1, fs, pass_type, f_range, filt_type="FIR",
+                            verbose=np.any([print_transitions, verbose]))
 
     # Remove any NaN on the edges of 'sig'
     sig, sig_nans = remove_nans(sig)
--- a/neurodsp/filt/iir.py
+++ b/neurodsp/filt/iir.py
@@ -1,5 +1,6 @@
 """Filter signals with IIR filters."""
 
+import numpy as np
 from scipy.signal import butter, sosfiltfilt
 
 from neurodsp.utils import remove_nans, restore_nans
@@ -10,8 +11,8 @@ from neurodsp.plts.filt import plot_frequency_response
 ###################################################################################################
 ###################################################################################################
 
-def filter_signal_iir(sig, fs, pass_type, f_range, butterworth_order,
-                      print_transitions=False, plot_properties=False, return_filter=False):
+def filter_signal_iir(sig, fs, pass_type, f_range, butterworth_order, print_transitions=False,
+                       plot_properties=False, return_filter=False, verbose=False):
     """Apply an IIR filter to a signal.
 
     Parameters
@@ -41,6 +42,8 @@ def filter_signal_iir(sig, fs, pass_type, f_range, butterworth_order,
         If True, plot the properties of the filter, including frequency response and/or kernel.
     return_filter : bool, optional, default: False
         If True, return the second order series coefficients of the IIR filter.
+    verbose : bool, optional, default: False
+        If True, print out detailed filter information.
 
     Returns
     -------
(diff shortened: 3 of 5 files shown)


TomDonoghue:
Hey @ryanhammonds - what are your thoughts on making this collect the info into a string that can be printed, as you have now, or saved out for later? Something like what FOOOF does for reports?

I'm thinking for the context of a script, or if there was a filter node in NDSPflow. One might want to keep a record of the filter details, but printing to console wouldn't necessarily be the easiest way to do so, and saving out the report might be useful. 

What do you think? If this would be super annoying, then it's probably not worth it, but if you think it wouldn't take too long, then it might be useful.

ryanhammonds:
Hi @TomDonoghue, I've added a `save_properties` kwarg to filter functions that saves a dictionary of filter properties as a json file. The new kwarg is a str pointing to the name of the json file to be saved. The `verbose` kwarg can still be used to print out filter properties as well.

ryanhammonds:
Also, tests are failing here do to an unrelated issue I think - nothing has changed in `sim.aperiodic` which are failing the tests. I think there was a weird bug in pytest that allowed that test to pass when ran with all of the others tests, but fails when running alone. Updating my version of pytest result in a fail everytime, even on master. I'll make a separate issue/PR adddressing this.

neurodsp/filt/utils.py
```diff
@@ -241,3 +256,96 @@ def remove_filter_edges(sig, filt_len):
...
+    Parameters
+    ----------
+    save_properties : str
+        Path, including file name, to save filter properites to as a json.
+    filt_report : dict
+        Contains filter report info.
+    """
+
+    # Ensure parents exists
+    if not os.path.isdir(os.path.dirname(save_properties)):
```
TomDonoghue, line 342: this directory check means that the save will fail if trying to save out to current directory (like: `save_properties='filt_report.json'`). I think we'd want this scenario to work.

neurodsp/filt/iir.py
```diff
@@ -39,8 +41,12 @@ def filter_signal_iir(sig, fs, pass_type, f_range, butterworth_order,
         If True, print out the transition and pass bandwidths.
     plot_properties : bool, optional, default: False
         If True, plot the properties of the filter, including frequency response and/or kernel.
+    save_properties : str
+        Path, including file name, to save filter properites to as a json.
```
TomDonoghue, line 45: typo on 'properites'

neurodsp/filt/fir.py
```diff
@@ -46,8 +47,12 @@ def filter_signal_fir(sig, fs, pass_type, f_range, n_cycles=3, n_seconds=None, r
         If True, print out the transition and pass bandwidths.
     plot_properties : bool, optional, default: False
         If True, plot the properties of the filter, including frequency response and/or kernel.
+    save_properties : str
+        Path, including file name, to save filter properites to as a json.
```
TomDonoghue, line 51: this file has same comments as iir:
- typo on this line
- any vs np.any on line 87

neurodsp/filt/utils.py
```diff
@@ -241,3 +256,96 @@ def remove_filter_edges(sig, filt_len):
...
+    filt_report['Filter Order'] = '{order}'.format(order=len(f_db)-1)
+
+    # Roll-off or transition bandwidth
+    filt_report['Transition Bandwidth'] = '{:.1f} Hz'.format(transition_bw)
+    filt_report['Pass/Stop Bandwidth'] = '{:.1f} Hz'.format(pass_bw)
+
+    # Passband ripple and stopband attenuation
+    pb_ripple = np.max(db[:np.where(f_db < f_range_trans[0])[0][-1]])
+    sb_atten = np.max(db[np.where(f_db > f_range_trans[1])[0][0]:])
+    filt_report['Passband Ripple'] = '{pb_ripple} db'.format(pb_ripple=pb_ripple)
```
TomDonoghue, line 309: should format these strings, to something like `:1.4f`

neurodsp/filt/utils.py
```diff
@@ -241,3 +256,96 @@ def remove_filter_edges(sig, filt_len):
...
+    filt_report['Stopband Attenuation'] = '{sb_atten} db'.format(sb_atten=sb_atten)
+
+    # Filter delay (zero-phase, linear-phase, non-linear phase)
+    filt_report['Filter Type'] = filt_type
+
+    if filt_type == 'FIR':
+
+        filt_report['Phase'] = '{filt_class}'.format(filt_class='linear-phase')
+        filt_report['Group Delay'] = '{delay}s'.format(delay=(len(f_db)-1) / (2 * fs))
+        filt_report['Direction'] = 'one-pass reverse'
```
TomDonoghue, line 319: why reverse? I'm not sure about that

neurodsp/filt/checks.py
```diff
@@ -164,18 +170,25 @@ def check_filter_properties(filter_coefs, a_vals, fs, pass_type, f_range,
...
              'pass/stop bandwidth of  {:.1f} Hz'.format(transition_bw, pass_bw))
 
-    # Print out transition bandwidth and pass bandwidth to the user
+    # Report filter properties
+    if verbose or save_properties:
+        filt_report = gen_filt_report(pass_type, filt_type, fs, f_db, db, pass_bw,
+                                      transition_bw, f_range, f_range_trans)
+
     if verbose:
-        print('Transition bandwidth is {:.1f} Hz.'.format(transition_bw))
```
TomDonoghue, line 177: I think, from the perspective of the `print_transitions` that is called from in the overlying filter functions, this shouldn't change (it should still just print the transitions, rather than the whole filter report). 
We could change the name here (verbose -> print_transitions) for consistency. 
This is technically breaking, but it's an internal function, so I think we can get away with it.

neurodsp/filt/utils.py
```diff
@@ -241,3 +256,96 @@ def remove_filter_edges(sig, filt_len):
...
+    filt_report = {}
+
+    # Filter type (high-pass, low-pass, band-pass, band-stop, FIR, IIR)
+    filt_report['Pass Type'] = '{pass_type}'.format(pass_type=pass_type)
+
+    # Cutoff frequenc(ies) (including definition)
+    filt_report['Cutoff (Half-Amplitude)'] = '{cutoff} Hz'.format(cutoff=f_range)
+
+    # Filter order (or length)
+    filt_report['Filter Order'] = '{order}'.format(order=len(f_db)-1)
```
TomDonoghue, line 300: order is something that is different between FIR & IIR. 
len - 1 is good for FIR. 
For IIR, the order we want to report is what is set as `butterworth_order` in `filter_signal_iir`

neurodsp/filt/utils.py
```diff
@@ -149,11 +152,15 @@ def compute_transition_band(f_db, db, low=-20, high=-3):
...
     high : float, optional, default: -3
         The upper limit that defines the transition band, in dB.
+    return_freqs : bool, optional, default: False
+        Whether to return a tuple of (lower, upper) frequency bounds for the transition band.
 
     Returns
     -------
     transition_band : float
         The transition bandwidth of the filter.
+    f_range : tuple of (float, float), optional, default: False
```
TomDonoghue, line 162: I don't think this should have `, optional, default: False`
A second line description (on 164) could / should say: "Only returned if `return_freqs` is True."

neurodsp/filt/utils.py
```diff
@@ -178,7 +185,15 @@ def compute_transition_band(f_db, db, low=-20, high=-3):
     # This gets the indices of transitions to the values in searched for range
     inds = np.where(np.diff(np.logical_and(db > low, db < high)))[0]
     # This steps through the indices, in pairs, selecting from the vector to select from
-    transition_band = np.max([(b - a) for a, b in zip(f_db[inds[0::2]], f_db[inds[1::2]])])
+    transition_pairs = [(a, b) for a, b in zip(f_db[inds[0::2]], f_db[inds[1::2]])]
```
TomDonoghue, line 188: the previous description here (line 187) is... not great. 
Since you've updated and extended this section, could you add some comments detailing more about what it's doing, and what it's extracting here?

TomDonoghue (changes requested):

TomDonoghue:
@ryanhammonds - okay, so I did a review, and left specific comments. 

More broadly, I still think we might want a slightly different approach here. Some thoughts:
- I think we should conserve the old function of `print_transitions`, printing only transition
    - when using filters on the fly, this info is useful
    - the whole filter report, by comparison, has a lot of useful information, but stuff that I generally know at the moment that I'm using the filter (ex, that I'm using an FIR bandpass), and so is not so useful to print in the moment, but useful to save out to a report for later. 
- I'm still voting for the 'report' to be more like what we do in FOOOF, meaning in my mind, collecting the information, and saving out a combined PDF that includes all the filter information, the plots (frequency response & kernel). 
- In terms of API, I think perhaps we dont add a 'verbose' argument. For printing info in real time, this can be done with `print_transitions` and/or `plot_properties`. I think the added argument, say `save_report` should collect this text and image output, and save out a PDF. 

Or, short version: let's keep `print_transitions` the same, make the filter report include the plots, and probably drop a separate 'verbose' argument. Does that make sense?

neurodsp/filt/iir.py
```diff
@@ -65,7 +71,9 @@ def filter_signal_iir(sig, fs, pass_type, f_range, butterworth_order,
     sos = design_iir_filter(fs, pass_type, f_range, butterworth_order)
 
     # Check filter properties: compute transition bandwidth & run checks
-    check_filter_properties(sos, None, fs, pass_type, f_range, verbose=print_transitions)
+    check_filter_properties(sos, None, fs, pass_type, f_range, filt_type="IIR",
+                            verbose=np.any([print_transitions, verbose]),
```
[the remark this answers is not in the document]
ryanhammonds: I'm dropping verbose as suggested.

neurodsp/filt/utils.py
```diff
@@ -241,3 +256,96 @@ def remove_filter_edges(sig, filt_len):
...
+        The transition bandwidth of the filter.
+    f_range : tuple of (float, float) or float
+        Cutoff frequency(ies) used for filter, specified as f_lo & f_hi.
+    f_range_trans : tuple of (float, float)
+        The lower and upper frequencies of the transition band.
+
+    Returns
+    -------
+    filt_report : dict
+        A dicionary of filter parameter keys and corresponding values.
```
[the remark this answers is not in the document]
ryanhammonds: Fixed

neurodsp/filt/utils.py
ryanhammonds: Fixed. In this cases, we'd want a `./` in front so I've written this in for relative pathing.

neurodsp/filt/iir.py
ryanhammonds: Fixed. I renamed this variable to `save_report` as suggested. I also changed the description to 'save a filter report'.

neurodsp/filt/utils.py
ryanhammonds: Yep, good idea!

neurodsp/filt/utils.py
ryanhammonds: I was confused about the terminology here after reading about it. When a filter is applied via np.convolve, I wasn't sure that is considered one-pass reverse, one-pass forward, or one-pass forward and reverse.

If this should be one-pass forward, that would mean this filter is also causal? Compared to two-pass (I think that's correct for IIR), which could be non-causal?

neurodsp/filt/checks.py
ryanhammonds: I reverted this back to how it was.

neurodsp/filt/utils.py
ryanhammonds: Thanks!

ryanhammonds:
@TomDonoghue I restore the functionality of `print_transitions`, removed `verbose`, and added `save_report` to save out plots and parameters as a pdf. I had to change almost everything in this PR, sorry! I thought we wanted something like the `fooof.report()` method and an additional json so a filter could be replicated easily.

ryanhammonds:
I met with Brad yesterday to discuss some of the things I am working on and he voted to keep `verbose` here, in addition to `save_report`. It can exists without conflicting with `print_transitions`  or `save_report`. I'll push this back in soon.

TomDonoghue:
So what will 'verbose' do then? It's not clear to me it adds any functionality...?

ryanhammonds:
It prints the filter report to console without having to open the pdf:
verbose

neurodsp/filt/utils.py
TomDonoghue: I think I would just say it's one pass? A one pass filter is not necessarily causal, which is a statement on whether future samples are used in the calculation. What we do is an acausal on-pass, I'm pretty sure (check out a kernel plot - the time definition is positive & negative time around the current sample). 

Also, I think we need to check the "group delay" we report because I think that's the delay if it were a causal filter. In practice we have a zero-phase filter, I believe (cus it's acausal). I might also be getting confused in the terminology here though. 

A two-pass (filt-filt) approach is always acausal, because the reverse pass uses "future" information.

TomDonoghue:
So yeh, in terms of what `verbose` would print out the informative info is available in `print_transition`. Things like "FIR" or (8-12) are available from the API call, and even info like "1-pass" are inherent in the filter called (will never change with different calls). If `verbose` is a shortcut for `print_transition` + `plot_properties` that seems... okay? But like, feels a bit like API clutter for not much benefit, and  with being `verbose` being such a common arg - I think it's not obvious that it would print plots. 

Anyways - sorry for the confusion on this one - referring back to your previous comment, I also have in mind that this should be like `fooof.report()` in terms of making that structured report - though I guess the difference is that in case it's not as much about printing in real time, but creating the report to save out. 

So - what still needs doing here?

ryanhammonds:
So the plots are only plotted in my example because ``plot_properties`` is passed. It's unrelated to ``verbose``, which only includes what is printed in the output similar to standard verbose args. There was a lot printed with verbose arg that isn't obvious (at least to me) and thought it may be useful.

Not a big deal to me either way so no need to go on about this. Verbose has been dropped already and I've fully rewritten this PR twice now. With the exception of the one unresolved comment, it should be how you wanted and ready to merge. Feel free to direct push any changes you want or to close.

neurodsp/filt/utils.py
ryanhammonds: I updated this to one-pass. I'm still a bit confused about how to determined direction based on scipy's firwin filter coefficients.

I updated the group delay to zero based on the Cheveigné 2019 paper: "It is delayed if the filter is causal. The delay may be avoided by choosing a zero-phase filter (green), but the response is then acausal."

PR #240, opened 2021-01-20
[FIX] - Update settings for knee test

TomDonoghue:
This is a follow up on #237, to fix up that update to make sure that the tests are consistent in using the same sampling rate for simulating the data as for computing the power spectrum. 

@ryanhammonds - this should be pretty procedural. Assuming this looks good to you on a quick glance, should be able to be merged in.


ryanhammonds (approved):
Looks good!

Issue #223, opened 2020-09-03
[BUG] sim_oscillation bug in power spectrum


ryanhammonds:
I was simulating sine waves and computing spectra:

```
from neurodsp.spectral import compute_spectrum
from neurodsp.plts import plot_time_series, plot_power_spectra
from neurodsp.utils import create_times
from neurodsp.sim import sim_oscillation

n_seconds = 10
fs = 1000

times = create_times(n_seconds, fs)
sig_8hz = sim_oscillation(n_seconds, fs, 8)
sig_9hz = sim_oscillation(n_seconds, fs, 9)

freqs_8hz, spec_8hz = compute_spectrum(sig_8hz, fs)
freqs_9hz, spec_9hz = compute_spectrum(sig_9hz, fs)

plot_power_spectra(freqs_8hz, spec_8hz, title='8hz')
plot_power_spectra(freqs_9hz, spec_9hz, title='9hz')
```
And notices strange spectra for odd freq waves:
![spec](https://user-images.githubusercontent.com/34786005/92167174-cb7a2300-edee-11ea-914f-38baf64e9322.png)

@rdgao Tracked this down to this line ``cycle = sim_cycle(n_seconds_cycle, fs, cycle, **cycle_params)`` in ``sim_oscillation`` and an interaction with sampling rate. When fs is increased to a large value, the odd spectra are corrected.

rdgao:
yup, and just adding a comment: 

I suspect this is because when an oscillation with a frequency that doesn't evenly divide the sampling rate (e.g., 9Hz and fs=1000Hz), our current approach creates non-perfect cycles of oscillations. They are either a little more than a cycle or a little less, and when you tile them to make the full signal, even if the signal length supports an integer number of cycles, as is the case here, you don't end up with the correct signal (the time series plots @ryanhammonds made were pretty diagnostic). 

Another way to say this: even though you can get 9 complete cycles of sine wave in 1 second sampled at 1000Hz, 1000 doesn't divide into 9, so depending on if you simulate all 9 cycles of 1 second (which will end at the correct place), or 1 cycle then tile it 9 times, the final signal will look different.

I suggested trying fs = 7200 which fixed the problem, because it's an integer multiple of both 8 and 9, so each individual cycle can be completed in an integer number of samples. I don't think this will work with arbitrary large sampling rates, in general.

rdgao:
I think this is related to #189

TomDonoghue:
[...]

PR #245, opened 2021-01-29
[ENH] Incomplete tile warning

ryanhammonds:
This adds a tile warning addressing #223, so noobs like myself aren't confused when a slope in the spectrum results from `sim_oscillation` (i.e. fs/freq is not an int). The slope below use to be smooth but now it produces a decreasing slope of harmonics.

```python


n_seconds = 100
fs = 1000
freq = 9

freqs, powers = compute_spectrum(sig, fs)

```

### DIFF
--- a/neurodsp/sim/periodic.py
+++ b/neurodsp/sim/periodic.py
@@ -40,10 +40,10 @@ def sim_oscillation(n_seconds, fs, freq, cycle='sine', phase=0, **cycle_params):
 
     Notes
     -----
-    When ``freq`` and ``fs`` are not evenly divisible, there will be a non-integer number of samples
-    per cycle. In the frequency domain, this will lead to power in non-simulated frequencies.
-    Consider updating ``freq and ``fs`` to an integer divisor/multiple when investigating spectral
-    properties.
+	Depending on the requested frequency (`freq`), sampling rate (`fs`), and signal
+	length (`n_seconds`), the simulated signal may have a non-integer number of cycles.
+	If so, the frequency-domain representation of the signal will have some power in non-simulated
+	frequencies. To avoid this, choose `n_seconds` and `fs` to create an integer number of cycles.
 
     Examples
     --------
--- a/neurodsp/sim/periodic.py
+++ b/neurodsp/sim/periodic.py
@@ -1,5 +1,6 @@
 """Simulating time series, with periodic activity."""
 
+import warnings
 from itertools import repeat
 
 import numpy as np
@@ -49,6 +50,9 @@ def sim_oscillation(n_seconds, fs, freq, cycle='sine', phase=0, **cycle_params):
     >>> sig = sim_oscillation(n_seconds=1, fs=500, freq=15,
     ...                       cycle='asine', phase=0.5, rdsym=0.75)
     """
+
+    # Check if tiling can produce full oscillations
+    _check_tiling(fs, freq)
 
     # Figure out how many cycles are needed for the signal
     n_cycles = int(np.ceil(n_seconds * freq))
@@ -149,6 +153,9 @@ def sim_bursty_oscillation(n_seconds, fs, freq, burst_def='prob', burst_params={
         temp = cycle_params.pop(burst_param, 0.2)
         if burst_def == 'prob' and burst_param not in burst_params:
             burst_params[burst_param] = temp
+
+    # Check if tiling can produce full oscillations
+    _check_tiling(fs, freq)
 
     # Simulate a normalized cycle to use for bursts
     n_seconds_cycle = 1/freq
@@ -303,3 +310,15 @@ def get_burst_samples(is_oscillating, fs, freq):
     bursts = np.repeat(is_oscillating, n_samples_cycle)
 
     return bursts
+
+
+def _check_tiling(fs, freq):
+    """Check if tiling can produce full oscillations."""
+
+    if not (fs/freq).is_integer():
+
+        warnings.warn('''
+        The requested frequency and sampling rate are not evenly divisible, resulting
+        in an artifactual spectral slope. Consider updating freq and fs to an integer
+        divisor/multiple pair.
+        ''', category=RuntimeWarning)
--- a/neurodsp/sim/periodic.py
+++ b/neurodsp/sim/periodic.py
@@ -1,6 +1,5 @@
 """Simulating time series, with periodic activity."""
 
-import warnings
 from itertools import repeat
 
 import numpy as np
@@ -39,6 +38,13 @@ def sim_oscillation(n_seconds, fs, freq, cycle='sine', phase=0, **cycle_params):
     sig : 1d array
         Simulated oscillation.
 
+    Notes
+    -----
+    When ``freq`` and ``fs`` are not evenly divisible, there will be a non-integer number of samples
+    per cycle. In the frequency domain, this will lead to power in non-simulated frequencies.
+    Consider updating ``freq and ``fs`` to an integer divisor/multiple when investigating spectral
+    properties.
+
     Examples
     --------
     Simulate a continuous sinusoidal oscillation at 5 Hz:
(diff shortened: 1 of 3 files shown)


neurodsp/sim/periodic.py
```diff
@@ -303,3 +310,15 @@ def get_burst_samples(is_oscillating, fs, freq):
...
+def _check_tiling(fs, freq):
+    """Check if tiling can produce full oscillations."""
+
+    if not (fs/freq).is_integer():
+
+        warnings.warn('''
+        The requested frequency and sampling rate are not evenly divisible, resulting
+        in an artifactual spectral slope. Consider updating freq and fs to an integer
+        divisor/multiple pair.
+        ''', category=RuntimeWarning)
```
TomDonoghue, line 324: I think we could perhaps add the middle step to this warning, which is that not evenly divisible -> uneven number of cycles -> a potentially unexpected power spectrum since I'm not sure mentioning an 'artifactual spectral slope' is obvious if one doesn't already know what's going on (and to nitpick - it's not strictly "artifactual" in that it is the correct / expected slope for a non integer number of cycles. 

Suggested re-wording:
"The settings for the frequency and sampling rate are not evenly divisible, meaning there is not an integer number of cycles for the requested simulated oscillation. In the frequency domain, this can lead to power in non-simulated frequencies in the power spectrum. You may want to consider updating freq and fs to an integer divisor/multiple pair."

^This is perhaps a bit long though, and could be condensed?

neurodsp/sim/periodic.py
```diff
@@ -303,3 +310,15 @@ def get_burst_samples(is_oscillating, fs, freq):
     bursts = np.repeat(is_oscillating, n_samples_cycle)
 
     return bursts
+
+
+def _check_tiling(fs, freq):
+    """Check if tiling can produce full oscillations."""
```
TomDonoghue, line 316: I'm not sure 'full oscillations' is totally obvious, so perhaps:
"Check if tiling will produce an integer number of cycles for the simulated oscillation."

TomDonoghue (changes requested):

TomDonoghue:
Very interesting idea!

I make some comments on the current warning to potential update the description. 

But then I was playing with it for a moment, and I realized it's actually more than fs & freq - it depends on the total simulated time as well. 
So far example, this would pass the test, but still lead to the 'weird slope':
```
n_seconds, fs, freq = 0.95, 1000, 10
freqs, powers = compute_spectrum(sig, fs)
```

Also: I'm a little wary that this warning might trigger a lot, and be annoying. 
Something like this is common:
```
for freq in [8, 9, 10, 11, 12]:
      sig = sim_oscillation(10, 1000, freq)
```
and it would be annoying to trigger the warning a ton. 
In this situation, the recommendation would typically not be use different sampling frequencies. Having non-integer number of cycles is something that happens sometimes - and sorta fine (it's not something to always avoid, and is not necessarily a big issue). 

So: maybe we need to update this warning to make it consider `n_seconds`. But even so, I'm a bit hesitant to have it trigger too often. Maybe this is something that should rather be discussed more in the tutorials? And/or, I think we should perhaps add the text proposed for the warning into the docstring, in a Notes section?

neurodsp/sim/periodic.py
ryanhammonds: Thanks! I like you re-wording - I'll condense it a bit.

ryanhammonds:
Thanks! I thought this was worth consideration since it seemed to spark a long discussion in #223.

I didn't realize this was dependent on `n_seconds` too. It looks your example is caused by `n_seconds` being too short to contain a single full cycle but I'll have to dig in more.

I can also see this warning being annoying - but I think this is important for spectral analyses? Ideally, it would be better if this was only triggered if `compute_spectrum` was ran on the simulated timeseries - but I don't think there's a way to do this without making things complicated.

I can drop the warning here and move it to notes.

ryanhammonds:
For your first example, scipy raises warning about defaulting `nperseg`. Your example can be fixed by passing in different values (`nperseg=900`). This only seems to be an issue when when `n_seconds < 1`.  I played with your second example and it only raised the warning once per notebook.

The latest commit moves the warning to a note in the sim_oscillation docstring.

TomDonoghue:
Hey @ryanhammonds :

Yeh, I also don't know that there is any way to trigger the warning from `compute_spectrum`. I don't think it's _that_ important - the amount of off-CF power is minimal, and note that this is something relevant to all PSDs (real data, for example, has no guarantee of an integer number of cycles, and a non-integer number leads to the same thing in the frequency domain). The scipy warning is if the default `nperseg` of 1 second is longer than the signal length, I think - which isn't that important. 

The `n_seconds` dependency is that if the duration of the signal doesn't cleanly include an integer number of cycles, then the signal gets truncated to end with a non-integer number of cycles, causing the same issue. (This technically happens when the sim'd signal is truncated at the end of sim_oscillation). 

Because of the n_seconds dependency, I've pushed a commit that suggests a updated phrasing of the note in the function. 

Also, it feels like even if this is not a continuously checked warning, then we should have functionality available to check this. To add this, I've added a sim util function to get an oscillation definition (in terms of whether the definition will include an integer number of cycles), which uses your original check for `fs`, and also check the signal duration. What do you think about this extra utility? If you like the idea, can you test it a little, to make sure if actually seems to track definitions that do / don't end up with integer number of cycles.

neurodsp/sim/utils.py
```diff
@@ -0,0 +1,30 @@
...
+    -------
+    bool
+        Whether the definition will have an integer number of cycles.
+    """
+
+    # Sampling rate check: check if the number of samples per cycle is an integer
+    srate_check = (fs/freq).is_integer()
+
+    # Time check: check if signal length matches an integer number of cycles
+    time_check = (n_seconds * freq).is_integer()
```
ryanhammonds, line 28: Is truncation always an issue? This example seems to produce an appropriate spectrum:
```python
n_seconds, fs, freq = 1.99, 1000, 10

print((n_seconds * freq).is_integer())

freqs, powers = compute_spectrum(sig, fs)
```

ryanhammonds (commented):
I like the idea of a separate util. I left one comment about truncation. This doesn't seem like it is always an issue and I'll have to run more sims to figure out when it is - so far though it seems like `n_seconds` is only an issue when < 1.

neurodsp/sim/utils.py
TomDonoghue: huh, great catch! 

So yeh, this time check isn't super specific. Whether the number of cycles leads to funky PSD is a bit more complex than this check - it depends on the actual spectral estimation that is done (for example, with Welchs, I think it depends on the segment size (`nperseg`), and how this relates to the signal, more than the details of the total signal. Because of this, I think we should probably just entirely drop this time check, since I otherwise don't really know how to make it more specific (we can't predict the spectral estimation people will do - and this emphasizes that the details of the resulting PSD are somewhat contextual - not fully determined by the sim'd time series).

TomDonoghue:
So, I feel like I should have realized this earlier - but an important thing to keep in mind here is that the power on/off frequency is not defined solely by the simulation, but also the manner of computing the power spectrum - in particular when we are talking about an integer number of cycles. 

Consider this sim:
```
from neurodsp.sim import sim_oscillation, sim_cycle

n_seconds, fs, freq = 10.0, 1000, 7

freqs, powers = compute_spectrum(sig, fs)
```

Gives the seemingly wonky power spectrum:
Screen Shot 2021-02-23 at 11 33 31 AM

However, this is all down to the relationship between the simulated frequency, and the `nperseg` value in the Welch's spectral estimation (and nothing inherent in the simulation). 

We can check this by updating nperseg:
```
cyclen = len(sim_cycle(1/freq, 1000, 'sine'))
freqs, powers = compute_spectrum(sig, fs, nperseg=cyclen*5)
```

Which gives the a more exact spectrum:
Screen Shot 2021-02-23 at 11 33 37 AM

By default, fs and nperseg are often multiples of 10, and so frequencies like 10 Hz divide evenly, and frequencies like 7 do not - but this isn't inherent to the simulation, and not predictable from the simulation definition (n_seconds, fs, freq), without knowing something about the spectral estimation. 

Which ultimately means - I don't think this 'helper' function or the note are actually well posed - since we can't say anything strictly true about the expected spectrum of the simulated signal based on the signal - what matters is the spectral estimation. So - in the end, I think we should drop this PR. 

@ryanhammonds - thoughts? Does this make sense? I don't currently see a clear way to integrate this into the codebase without it being heavily caveated. As an alternative, this is all some information that I think would be useful to work into a tutorial / example somewhere - and if you feel you have a handle on it, a quick example that discusses and shows the relationship between spectral estimation and expected power spectra would be quite useful I think.

PR #246, opened 2021-02-02
[ENH] 2d Array Plotting

ryanhammonds:
This updates `plot_time_series` to accept 2d arrays, in addition to a list of 1d arrays. Before this would not work:

```python
import numpy as np

n_sigs = 10
fs = 1000
n_seconds = 10

sigs = np.repeat(np.arange(n_sigs), int(n_seconds * fs))
sigs = sigs.reshape((n_sigs, int(n_seconds * fs)))

times = np.arange(0, len(sigs[0])/fs, 1/fs)

plot_time_series(times, sigs)
```

and it would require:
```python
plot_time_series(times, list(sigs))
```

### DIFF
--- a/neurodsp/plts/time_series.py
+++ b/neurodsp/plts/time_series.py
@@ -19,9 +19,9 @@ def plot_time_series(times, sigs, labels=None, colors=None, ax=None, **kwargs):
 
     Parameters
     ----------
-    times : 1d array or list of 1d array
+    times : 1d array, list of 1d array, or 2d array
         Time definition(s) for the time series to be plotted.
-    sigs : 1d array or list of 1d array
+    sigs : 1d array, list of 1d array, or 2d array
         Time series to plot.
     labels : list of str, optional
         Labels for each time series.
@@ -47,18 +47,24 @@ def plot_time_series(times, sigs, labels=None, colors=None, ax=None, **kwargs):
 
     ax = check_ax(ax, (15, 3))
 
-    times = repeat(times) if isinstance(times, np.ndarray) else times
-    sigs = [sigs] if isinstance(sigs, np.ndarray) else sigs
+    n_repeats = len(sigs) if isinstance(sigs, list) or sigs.ndim == 2 else 1
+
+    # Repeat times if needed
+    if isinstance(times, np.ndarray) and times.ndim != 2:
+        times = np.tile(times, (n_repeats, 1))
+
+    # Make sigs iterable if 1D
+    sigs = np.reshape(sigs, (1, -1)) if not isinstance(sigs, list) and sigs.ndim == 1 else sigs
 
     if labels is not None:
         labels = [labels] if not isinstance(labels, list) else labels
     else:
-        labels = repeat(labels)
+        labels = repeat(labels, n_repeats)
 
     # If not provided, default colors for up to two signals to be black & red
     if not colors and len(sigs) <= 2:
         colors = ['k', 'r']
-    colors = repeat(colors) if not isinstance(colors, list) else cycle(colors)
+    colors = repeat(colors, n_repeats) if not isinstance(colors, list) else cycle(colors)
 
     for time, sig, color, label in zip(times, sigs, colors, labels):
         ax.plot(time, sig, color=color, label=label)
--- a/neurodsp/plts/time_series.py
+++ b/neurodsp/plts/time_series.py
@@ -1,6 +1,6 @@
 """Plots for time series."""
 
-from itertools import cycle
+from itertools import repeat, cycle
 
 import numpy as np
 import numpy.ma as ma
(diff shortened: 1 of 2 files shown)


neurodsp/plts/time_series.py
```diff
@@ -19,9 +19,9 @@ def plot_time_series(times, sigs, labels=None, colors=None, ax=None, **kwargs):
 
     Parameters
     ----------
-    times : 1d array or list of 1d array
+    times : 1d array, list of 1d array, or 2d array
         Time definition(s) for the time series to be plotted.
-    sigs : 1d array or list of 1d array
+    sigs : 1d array, list of 1d array, or 2d array
```
TomDonoghue, line 24: probably this can be condense to say something like:
`1d or 2d array or list of 1d array`
I'm not sure if there's an exact convention for specifying a range of array sizes - but one way or another the array_type input can probably be grouped, and same for the `times` input

neurodsp/plts/time_series.py
```diff
@@ -47,18 +47,24 @@ def plot_time_series(times, sigs, labels=None, colors=None, ax=None, **kwargs):
 
     ax = check_ax(ax, (15, 3))
 
-    times = repeat(times) if isinstance(times, np.ndarray) else times
```
TomDonoghue, line 50: if `times` is a 2d array, then it can be iterated by leaving it as is, without needing the tile in lines 53-54, right?
In this case, this line just needs to be updated to consider dimensionality, such as:
`times = repeat(times) if (isinstance(arr, np.ndarray) and arr.ndim == 1) else times`

neurodsp/plts/time_series.py
```diff
@@ -47,18 +47,24 @@ def plot_time_series(times, sigs, labels=None, colors=None, ax=None, **kwargs):
...
-    sigs = [sigs] if isinstance(sigs, np.ndarray) else sigs
```
TomDonoghue, line 51: basically the same update as the `times`: if this line considers dimensionality, then the reshaping in 57 isn't needed:
sigs = [sigs] if (isinstance(arr, np.ndarray) and arr.ndim == 1) else sigs

neurodsp/plts/time_series.py
```diff
@@ -47,18 +47,24 @@ def plot_time_series(times, sigs, labels=None, colors=None, ax=None, **kwargs):
...
+    n_repeats = len(sigs) if isinstance(sigs, list) or sigs.ndim == 2 else 1
```
TomDonoghue, line 50: what is the goal / benefit of making an explicit number of repeats, for colors / labels / times. 
The idea before was that the limiting factor for the plot loop should be `sigs` - it will loop for the number of signals available. The benefit of _not_ making an explicit number of repeats, is that if I want to plot every second signal in red or something, I could pass 4 signals with `colors=['black', 'red']`, which would no longer work. Because of this, I'm not sure I see any reason for the explicit repeats, and would probably say step that back?

TomDonoghue (changes requested):

TomDonoghue:
Yeh, this is a totally sensible update. I've made some review comments for condensing etc. Also: maybe extend the test to test the case of having a 2d array input? Also, there are some 'secondary' functions which use `plot_time_series` (such as `plot_bursts` and `plot_instantaneous_measure` - can you have a quick check that this update doesn't induce anything weird with their outputs?

Here is a slightly different example in which this update does help to plot things cleanly:
```

n_seconds, fs = 2, 1000

powers = [0, 0.25, 0.5, 0.75, 1]

sigs = np.array([sim_oscillation(n_seconds, fs, 10, variance=power) for power in powers])

plot_time_series(times, sigs)
```
Output:
Screen Shot 2021-02-05 at 12 50 25 AM

This is a slightly different from the notion of plotting something like 'multiple channels':
```

n_seconds, fs = 2, 1000
powers = [0, 0.25, 0.5, 0.75, 1]

sigs = np.array([sim_oscillation(n_seconds, fs, 10, variance=power) + 2*ind for ind, power in enumerate(powers)])

plot_time_series(times, sigs)
```

output: 
Screen Shot 2021-02-05 at 12 54 44 AM

Note, for example, that the y-axis labels and ticks don't really make sense in this case. 

The point being: I think we should merge this update in for the current function, but this also perhaps highlights that we should add a _different_ / new function for doing 'multi-channel'?

ryanhammonds:
Thanks for the review! I went a little wild with trying to avoid to use repeat. I've incorporated your feedback.

This plays nicely with `plot_instantaneous_measure`. And doesn't change anything with `plot_bursts` - it only took a 1d array and that hasn't changed. A list or 2d array will break `plot_bursts`. I could change that if you thinks it's useful. I'll take some playing around with since vstacking sig and bursts doesn't work.

I think a multi-channel plotting function would be useful and wouldn't be too much work to write in but might overlap with MNE. I took a stab at this in ndspflow with plotly - it ended up not working super well for large arrays there since there was too much interactivity and something deep in plotly's codebase/dependencies caused slow loading and poor performance.

TomDonoghue (approved):

TomDonoghue:
If this doesn't change anything with instaneous, and bursts, I think we're good. I don't think it's an issue that bursts doesn't accept 2d - it's not even clear what it should do in that case, and it's documented to only take in a 1d array. 

With that, the code & update in this PR all looks good to me, you can merge it in @ryanhammonds!

In terms of the multi-channel plotting (as a potential separate thing, not for this PR) - I agree, it could be useful, and try it out if you want - but it's not a strong priority necessarily.

PR #248, opened 2021-02-18
[ENH] - Add support for passing in a callable to sim_cycle

TomDonoghue:
This update allows for passing in a callable to `sim_cycle`. 

This is to make it easier to explore new cycle types. In particular, this means a callable can be passed through from `sim_oscillation`. I think this useful to be able to be able to explore and test new cycles and tiling, without having to add them to the module. If the function is well formed as a cycle function, it gets called, with any passed in parameters. 

Consider the following example that will now work:
```
def sim_offset_sine(n_seconds, fs, offset):

    cycle = np.sin(2 * np.pi * 1 / n_seconds * (np.arange(int(fs * n_seconds)) / fs))
    cycle = cycle + offset

    return cycle

n_seconds, fs, freq = 1, 1000, 10
osc = sim_oscillation(n_seconds, fs, freq, sim_offset_sine, offset=1, mean=None)
# Note: passing in the `mean` param is just to turn off normalization.
```

### DIFF
--- a/neurodsp/sim/cycles.py
+++ b/neurodsp/sim/cycles.py
@@ -21,7 +21,7 @@ def sim_cycle(n_seconds, fs, cycle_type, **cycle_params):
         This is NOT the period of the cycle, but the length of the returned array of the cycle.
     fs : float
         Sampling frequency of the cycle simulation.
-    cycle_type : {'sine', 'asine', 'sawtooth', 'gaussian', 'exp', '2exp'}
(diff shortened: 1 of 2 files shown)


ryanhammonds (commented):
Looks good! The only thing I could think of was updating the docstring to note the new functionality.

TomDonoghue:
Thanks for the quick check @ryanhammonds! I've added a quick update to the docstrings noting this update.

PR #252, opened 2021-02-24
[ENH] Simulate variable oscillations

ryanhammonds:
This allows the frequency and/or simulation kwarg(s) to vary on a cycle-by-cycle basis.

```python
import numpy as np
from neurodsp.sim import sim_variable_oscillation

fs = 1000

freqs = [ 5, 10, 15, 20]
rdsyms= [.2, .4, .6, .8]

sig = sim_variable_oscillation(fs, freqs, cycle='asine', rdsym=rdsyms)
times = np.arange(0, len(sig)/fs, 1/fs)

plot_time_series(times, sig)
```
Screenshot_20210223_173332

### DIFF
--- a/neurodsp/sim/periodic.py
+++ b/neurodsp/sim/periodic.py
@@ -180,19 +180,19 @@ def sim_bursty_oscillation(n_seconds, fs, freq, burst_def='prob', burst_params=N
 
 
 @normalize
-def sim_variable_oscillation(fs, freqs, n_seconds='auto', cycle='sine', phase=0, **cycle_params):
+def sim_variable_oscillation(n_seconds, fs, freqs, cycle='sine', phase=0, **cycle_params):
     """Simulate an oscillation that varies in frequency and cycle parameters.
 
     Parameters
     ----------
+    n_seconds : float or None
+        Simulation time, in seconds. If None, the simulation time is based on `freqs` and the
+        length of `cycle_params`. If a float, the signal may be truncated or contain trailing zeros
+        if not exact.
     fs : float
         Signal sampling rate, in Hz.
     freqs : float or list
         Oscillation frequencies.
-    n_seconds : 'auto' or float, optional, default: 'auto'
-        Simulation time, in seconds. If 'auto', the simulation time is based on `freqs` and the
-        length of `cycle_params`. If a float, the signal may be truncated or contain trailing zeros
-        if not exact.
     cycle : {'sine', 'asine', 'sawtooth', 'gaussian', 'exp', '2exp'} or callable
         Type of oscillation cycle to simulate.
         See `sim_cycle` for details on cycle types and parameters.
@@ -250,12 +250,12 @@ def sim_variable_oscillation(fs, freqs, n_seconds='auto', cycle='sine', phase=0,
     cycle_params = [{}] * len(freqs) if len(cycle_params) == 0 else cycle_params
 
     # Determine start/end indices
-    cyc_lens = [len(create_cycle_time((1 / freq), fs)) for freq in freqs]
+    cyc_lens = [int(np.ceil(1 / freq * fs)) for freq in freqs]
     ends = np.cumsum(cyc_lens, dtype=int)
     starts = [0, *ends[:-1]]
 
     # Simulate
-    n_samples = np.sum(cyc_lens) if n_seconds == 'auto' else int(n_seconds * fs)
+    n_samples = np.sum(cyc_lens) if n_seconds is None else compute_nsamples(n_seconds, fs)
 
     sig = np.zeros(n_samples)
 
--- a/neurodsp/sim/periodic.py
+++ b/neurodsp/sim/periodic.py
@@ -8,8 +8,8 @@ from neurodsp.utils.norm import normalize_sig
 from neurodsp.utils.data import compute_nsamples
 from neurodsp.utils.checks import check_param_range
 from neurodsp.utils.decorators import normalize
-from neurodsp.sim.cycles import sim_cycle, sim_normalized_cycle, phase_shift_cycle
-
+from neurodsp.sim.cycles import (sim_cycle, sim_normalized_cycle,
+                                 phase_shift_cycle, create_cycle_time)
 ###################################################################################################
 ###################################################################################################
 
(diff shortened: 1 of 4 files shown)


TomDonoghue:
I think this is a super cool / super useful extension - thanks @ryanhammonds!

One general thought: this function doesn't have quite the same API as our general sim functions, as it doesn't have `n_seconds` as a first input. Obviously it doesn't strictly need it - but this does mean this function won't naturally work in `sim_combined`, for example. 
What do you think about adding `n_seconds` as a first parameter? This input could be given as None, in which case the function would proceeds as it is currently. If it's not None, the function does something like initializing to the desired length, and filling in cycles for that signal length. In this case, the user might have to keep in mind the relationship between the given parameters, and the output length, but the benefit is that this function could be plugged into `sim_combined` and proceed as normal from there.

Also, a potential extension (not necessarily for this PR):
In spectral param, we have generators that help with sampling simulated parameters from a set, or stepping across them systematically. That might be useful here, for being able to sample some variability for given parameters, without pre-specifying every value. I mention it now, as it might help motivate adding `n_seconds` as well - since it might be useful to simulate an `n_second` signal, sampling variation of (for example) the rdsym for each cycle. If this sounds interesting, then I think the main update would be to (now or sometime later) add support for generators to this approach.

ryanhammonds:
I updated this to take an optional `n_seconds` argument so it would work with `sim_combined`.

```python
from neurodsp.sim import sim_variable_oscillation, sim_combined
import numpy as np

fs = 1000

freqs = [ 5, 10, 15, 20]
rdsyms= [.2, .4, .6, .8]

n_seconds = np.sum(1 / np.array(freqs))

components = {'sim_variable_oscillation':
                 {'freqs': freqs,
                  'cycle': 'asine',
                  'rdsym': rdsyms},
              'sim_powerlaw': {'exponent': -2}
             }
    
sig = sim_variable_oscillation(fs, freqs, cycle='asine', rdsym=rdsyms)
times = np.arange(0, len(sig)/fs, 1/fs)
plot_time_series(times, sig)


sig = sim_combined(n_seconds, fs, components, component_variances=[0.9, 0.05])
times = np.arange(0, len(sig)/fs, 1/fs)
plot_time_series(times, sig)
```
ex

ryanhammonds:
I rebased this to be compatible with main. There were some cycle edge artifacts that are now fixed too.

```python
import numpy as np

from neurodsp.sim import sim_variable_oscillation

# Slightly vary each cycle's freq
freqs = [10 + np.random.choice(np.arange(0, 4.25, .25)) for _ in range(10)]
fs = 1000
sig = sim_variable_oscillation(fs, freqs)

times = np.arange(0, len(sig)/fs, 1/fs)
plot_time_series(times, sig)
```

neurodsp/sim/periodic.py
```diff
@@ -179,6 +179,97 @@ def sim_bursty_oscillation(n_seconds, fs, freq, burst_def='prob', burst_params=N
...
+def sim_variable_oscillation(fs, freqs, n_seconds='auto', cycle='sine', phase=0, **cycle_params):
+    """Simulate an oscillation that varies in frequency and cycle parameters.
+
+    Parameters
+    ----------
+    fs : float
+        Signal sampling rate, in Hz.
+    freqs : float or list
+        Oscillation frequencies.
+    n_seconds : 'auto' or float, optional, default: 'auto'
```
TomDonoghue, line 192: I think I would vote for using 'None', instead of the string 'auto' here, for the in which `n_seconds` is to be inferred from frequency parameters. I think this would be consistent with other code (if value is not given, infer it from other params). This stops the param being mixed type - and I think should general work with the code (?).

neurodsp/sim/periodic.py
```diff
@@ -179,6 +179,97 @@ def sim_bursty_oscillation(n_seconds, fs, freq, burst_def='prob', burst_params=N
...
+            param_values[idx] = [param] * n_cycles
+
+    param_values = np.array(param_values).transpose()
+
+    # Collect params for each cycle separately
+    cycle_params = [dict(zip(param_keys, params)) for params in param_values]
+    cycle_params = [{}] * len(freqs) if len(cycle_params) == 0 else cycle_params
+
+    # Determine start/end indices
+    cyc_lens = [len(create_cycle_time((1 / freq), fs)) for freq in freqs]
```
TomDonoghue, line 253: I don't think the `create_cycle_time` is strictly needed here - it's making whole arrays, when what we really need is the length. Within `create_cycle_time`, the length comes from `np.arange(n_seconds, fs)`, which ends up being equivalent to creating an range from 0 to `int(np.ceil(n_seconds, fs))` in integer steps. With this, we don't need to generate the array, as the length of this array is equivalent to `int(np.ceil(n_seconds, fs))`. 

Long story short - I think this line can be written as:
`cyc_lens = [int(np.ceil(1 / freq * fs)) for freq in freqs]`
This is a little more efficient, and saves an import. 

If you want to play with this - I basically checked that nothing diverges in this code block to convince myself:
```
n_seconds, fs = 0.234, 101
print(n_seconds * fs)
print(int(np.ceil(n_seconds * fs)))
print(len(create_cycle_time(n_seconds, fs)))
print(len(np.arange(n_seconds * fs)))
```

neurodsp/sim/periodic.py
```diff
@@ -179,6 +179,97 @@ def sim_bursty_oscillation(n_seconds, fs, freq, burst_def='prob', burst_params=N
...
+    ends = np.cumsum(cyc_lens, dtype=int)
+    starts = [0, *ends[:-1]]
+
+    # Simulate
+    n_samples = np.sum(cyc_lens) if n_seconds == 'auto' else int(n_seconds * fs)
```
TomDonoghue, line 258: I think computing n_samples at the end here should use `compute_nsamples`, imported from `neurodsp.utils.data`. That function defines our convention for computing n_samples, which includes using a ceil. Using this import makes our computed number of samples consistent, for combining in `sim_combined` (I've run into bugs with getting different number of samples between signal segments before, and I think the easiest is to always use the same function to compute the expected number).

TomDonoghue:
Okay, I'm finally getting back to this!

I left a couple small code comments. Overall, I think this is a really cool addition!

I still like the idea of adding generator support one day, but that's a task for another day. 
One thing I want to chat on quick @ryanhammonds is to check in on thought or param ordering. But otherwise, after the current code comments, this looks good to me!

TomDonoghue:
Discussion points:
- Move `n_seconds` to first position (match other functions), making `n_seconds` required - but if you pass in None, it will be inferred (as it does with 'auto'). 
- Extension - can we add cycle by cycle variance scaling: if we add a `cycle_variances` input, and use `sim_normalized_cycle`, then we can support the option of scaling each cycle independently.

ryanhammonds:
@TomDonoghue I addressed everything here. The one difference is not using a separate kwarg for `cycle_variances` to keep things a bit more simple. Instead, `variance` can be a single value or a list. If a single value each cycle gets normalized to the same mean and/or variance, which should be the same thing as normalizing the cycle as a whole.

```python
import numpy as np
from neurodsp.sim import sim_variable_oscillation

fs = 1000
freqs = [ 5, 10, 15, 20]
rdsyms= [.2, .4, .6, .8]

# Single norm
sig = sim_variable_oscillation(None, fs, freqs, cycle='asine', rdsym=rdsyms, variance=10, mean=5)
assert np.mean(sig) == 5
assert np.var(sig) == 10

# Mutli norm
sig = sim_variable_oscillation(None, fs, freqs, cycle='asine', rdsym=rdsyms, variance=[0, 10, 0, 10])
times = np.arange(0, len(sig)/fs, 1/fs)
plot_time_series(times, sig)
```

TomDonoghue:
All looks good here - super cool update @ryanhammonds! Merging in!

PR #253, opened 2021-03-01
[MNT] - Use cumsum from numpy instead of scipy

TomDonoghue:
I noticed a deprecation warning that `scipy.cumsum` is being deprecated, and will be removed in scipy 2.0 - so this updates to use cumsum from numpy instead, as recommended by the scipy deprecation. The results should be identical between the two functions, AFAIK.


ryanhammonds (approved):
Looks good!

PR #254, opened 2021-03-01
[ENH] - Add one sided asymmetric option to asine cycle

TomDonoghue:
This update adds support for creating an asymmetric cycle with only one asymmetric extrema, to `sim_asine_cycle`.

For example, this code:
```
from neurodsp.sim.cycles import sim_asine_cycle

fs, freq = 1000, 10
cycle = sim_asine_cycle(1/freq, fs, 0.25, 'peak')

plt.plot(cycle)
```

Would now work, to have a 'peak-only asymmetric cycle', and gives this:
Screen Shot 2021-03-01 at 8 42 46 AM

As a limitation of this - it seems unideal that there is an abrupt shift from the asymmetric to symmetric regime at the zero crossing point (ideally this might be smoother), but broadly I think this should be _okay_, and this update can still be useful for some things. 

Edit: I've been playing around with those code for awhile, including poking at different ways to do the offsets, and if there's a way to refactor what looks to be duplicated code. At this point, I _think_ it's probably easiest to stick with it as is, but one way or another I'm moving this to review stage to get some more eyes on it!

### DIFF
--- a/neurodsp/sim/cycles.py
+++ b/neurodsp/sim/cycles.py
@@ -111,7 +111,7 @@ def sim_sine_cycle(n_seconds, fs):
     return cycle
 
 
-def sim_asine_cycle(n_seconds, fs, rdsym):
+def sim_asine_cycle(n_seconds, fs, rdsym, side='both'):
     """Simulate a cycle of an asymmetric sine wave.
 
     Parameters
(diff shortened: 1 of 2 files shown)


ryanhammonds (approved):
Everything looks good here in terms of functionality. I tested this out in a note book but am still wrapping my head around how the various phase arrays.

TomDonoghue:
Thanks for checking @ryanhammonds! I don't think anything necessarily needs any additional figuring out here, so I'm going to call it, and merge this one in!

PR #255, opened 2021-03-14
[MNT] - Update tests & small plot bug fixes

TomDonoghue:
This started as a quick bug fix for a plot thing, and then escalated a little....

Updates in this PR:
- Bug fix: fixes labelling on a phase instantaneous frequency so it plays more nicely with subplots
- Makes test plots more realistic, so that saved out plots are more representative and easier to check
- Simplifies test settings and fixtures**

**We were starting to get a slightly unwieldy collection of settings and fixtures. As I was trying to generalize the plots, using more realistic signals, it became clear the mapping between test signals and settings was not super clear. So, a quick clean up which basically simplifies test signals to have one regime of (N_SECONDS, FS) for fixtures defined in conftest, and then makes any tests that need idiosyncratic custom test object definitions create their own local simulations. 

Note: the changes here should be relatively simple, even if the diff is somewhat large. Should only need a relatively quick check.

### DIFF
--- a/neurodsp/plts/time_series.py
+++ b/neurodsp/plts/time_series.py
@@ -147,7 +147,5 @@ def plot_bursts(times, sig, bursting, ax=None, **kwargs):
     >>> plot_bursts(times, sig, is_burst, labels=['Raw Data', 'Detected Bursts'])
     """
 
-    ax = check_ax(ax, (15, 3))
-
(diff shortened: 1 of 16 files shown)


ryanhammonds (approved):
Everything looks good here!

TomDonoghue:
Thanks for the check!

PR #256, opened 2021-03-14
[BUG] - Fix bug in `sim_bursty_oscillation`

TomDonoghue:
There was a bug in `sim_bursty_oscillation`, due to defining an empty dictionary in the function definition, which turns out to be quite bad (ooops, my bad). Because we dynamically update the dictionary in the function, this updates the default value, leading to issues. 

Here's an isolated example of the issue:
```
def test_dict_issue(var={}):
    if not var:
        var['new'] = 'value'
        
print(test_dict_issue.__defaults__)
test_dict_issue()
print(test_dict_issue.__defaults__)
```
Which gives this output
```
({},)
({'new': 'value'},)
```

Basically, giving an empty dictionary as default, and then updating this object updates the object linked in the function definition. Once I tracked this down - it's a simple fix, but it's a bit of a funky bug. 

Anyways, interesting / weird Python quick. I sorta knew defaulting like this was bad, but now I actually get it - so lesson learned!

### DIFF
--- a/neurodsp/sim/periodic.py
+++ b/neurodsp/sim/periodic.py
@@ -68,7 +68,7 @@ def sim_oscillation(n_seconds, fs, freq, cycle='sine', phase=0, **cycle_params):
     return sig
 
 
-def sim_bursty_oscillation(n_seconds, fs, freq, burst_def='prob', burst_params={},
+def sim_bursty_oscillation(n_seconds, fs, freq, burst_def='prob', burst_params=None,
                            cycle='sine', phase=0, **cycle_params):
     """Simulate a bursty oscillation.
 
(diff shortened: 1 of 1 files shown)


ryanhammonds (approved):
Good catch! I didn't know about this. It's weird that variables aren't defaulted each time the function is called.

TomDonoghue:
Yeh - it's a bit of a weird one. I'd seen it reference in like linting notes. I think it's basically that the function definition gets compiled, with a reference to any default values / objects - so if the referant changes, so does the value accessed at run time. Basically, I think it has something do to with how function definitions get executed. 

Anyways - thanks for checking! I'll merge this in!

PR #258, opened 2021-04-15
[ENH] - Update default for savefig

TomDonoghue:
This updates the `savefig` decorator to default to saving a figure when a file name is given, meaning `save_fig` does not have to be explicitly passed. The `save_fig` input can still be explicitly passed to stop the figure from saving. 

Following this update, this will now save:
`example_plot(file_path=TEST_PLOTS_PATH, file_name='test_savefig1.pdf')`

This will now not save:
`example_plot(save_fig=False, file_path=TEST_PLOTS_PATH, file_name='test_savefig3.pdf')`
^This might seem like a slightly weird construct, but it's something I use a fair amount - to define the name I need, but use a general setting to define when I actually save things.

This PR also updates the tests, generalizing them to check this new behaviour. It also moves away from using tempfile for these particular files, since that doesn't work well in this case (since the tempfiles are defined, the path check passes whether or not the figure file is saved).

### DIFF
--- a/neurodsp/plts/utils.py
+++ b/neurodsp/plts/utils.py
@@ -34,9 +34,13 @@ def savefig(func):
     @wraps(func)
     def decorated(*args, **kwargs):
 
-        save_fig = kwargs.pop('save_fig', False)
+        # Grab file name and path arguments, if they are in kwargs
         file_name = kwargs.pop('file_name', None)
         file_path = kwargs.pop('file_path', None)
+
+        # Check for an explicit argument for whether to save figure or not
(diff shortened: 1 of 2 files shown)


ryanhammonds (approved):
Looks good! I confirmed that `file_name` can include an absolute/relative path. Sorry for the over complicated PR in fooof.

TomDonoghue:
Thanks for the review! And no worries at all - exploring code ideas is all part of the game, and helps us move towards the solutions we decide to merge :)

PR #259, opened 2021-04-20
[MRG] Simulate spikes / action potentials

ryanhammonds:
This adds action potential cycles as the sum of two inverted asymmetrical gaussians:

```python
import numpy as np
from neurodsp.sim.cycles import sim_ap_cycle

n_seconds = 1
fs = 1000

cycle_params = {'centers': (.25, .5), 'stds':(.15, .12),
                'alphas':(4, .2), 'heights': (15, -5)}

cyc = sim_ap_cycle(n_seconds, fs, **cycle_params)

plt.plot(cyc)
```
ap_cycle

### DIFF
--- a/neurodsp/sim/cycles.py
+++ b/neurodsp/sim/cycles.py
@@ -1,7 +1,8 @@
 """Simulating individual cycles, of different types."""
 
 import numpy as np
-from scipy.signal import gaussian, sawtooth
+from scipy.signal import sawtooth
+from scipy.stats import norm
 
 from neurodsp.sim.info import get_sim_func
 from neurodsp.utils.data import compute_nsamples
@@ -22,15 +23,18 @@ def sim_cycle(n_seconds, fs, cycle_type, phase=0, **cycle_params):
         This is NOT the period of the cycle, but the length of the returned array of the cycle.
     fs : float
         Sampling frequency of the cycle simulation.
-    cycle_type : {'sine', 'asine', 'sawtooth', 'gaussian', 'exp', '2exp'} or callable
+    cycle_type : {'sine', 'asine', 'sawtooth', 'gaussian', 'exp', '2exp', 'ap'} or callable
         What type of cycle to simulate. Options:
 
         * sine: a sine wave cycle
         * asine: an asymmetric sine wave
         * sawtooth: a sawtooth wave
         * gaussian: a gaussian cycle
+        * skewed_gaussian: a skewed gaussian cycle
         * exp: a cycle with exponential decay
         * 2exp: a cycle with exponential rise and decay
+        * ap: an action potential
+
 
     phase : float or {'min', 'max'}, optional, default: 0
         If non-zero, applies a phase shift by rotating the cycle.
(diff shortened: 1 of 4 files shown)


ryanhammonds:
This func can be used with scipys curve_fit to fit  segmented APs from real data (bycycle-tools/bycycle#100). The resulting 6 params could be useful (i.e. dimensionality reduction for APs) and included in df_features returned from bycycle.Spikes.fit().
ap_cycle_fit

neurodsp/sim/cycles.py
```diff
@@ -243,7 +253,89 @@ def sim_gaussian_cycle(n_seconds, fs, std):
...
     """
 
-    cycle = gaussian(compute_nsamples(n_seconds, fs), std * fs)
+    xs = np.linspace(0, 1, compute_nsamples(n_seconds, fs))
+    cycle = height * np.exp(-(xs-center)**2 / (2*std**2))
+
+    return cycle
+
+
+def sim_skewed_gaussian_cycle(n_seconds, fs, center, std, alpha, height=1):
```
TomDonoghue, line 262: The parameter `alpha` isn't described in the docstring

TomDonoghue (changes requested):

neurodsp/sim/cycles.py
```diff
@@ -230,6 +236,10 @@ def sim_gaussian_cycle(n_seconds, fs, std):
         Sampling frequency of the cycle simulation.
     std : float
         Standard deviation of the gaussian kernel, in seconds.
+    center : float, optional, default: 0.5
+        The center of the gaussian.
```
TomDonoghue, line 240: This definition isn't super clear (it's not obvious if it's about x or y axis), and I think the description in `sim_skewed_gaussian` is clearer and could be copied here.

TomDonoghue:
This looks super cool! 

I had a quick scan of the code, left a coupe comments, and here's a couple quick "higher-level" questions:
- It's not clear to me that the AP sim is a _cycle_, per se. Would it fit better in `sim.transients`?
- I presume changing the gaussian code to not use scipy is about getting more control over height, center, etc? Setting `height` feels like it overlaps with normalizing the signal, and `center` feels a little weird to have in the cycles, because it makes looks like it 'cuts off' the edge of a gaussian? Am I right in thinking that one would probably never want to set `center` to a custom value because it would create weird concatenation artifacts? 

Between the two comments, I think it's possible it might be a cleaner split to put some stuff in transients, that might work well for generating APs, but might not be quite the same as what one wants for a cycle? What do you think?

ryanhammonds:
Thanks for the review!

- I think it makes sense to refactor into `transients` with an alias in `cycles` so it can still be used with sim_bursty_oscillation.
- The main reason for the sim_gaussian_cycle changes was to have control over the cycle center, which is important only for skewed cycles. Here's an example of how center makes a difference with skewed cycles:

skew_center

The `height` param is a bit different from `variance`. I'm not sure there is a clear mapping between these params.
variance_v_height

ryanhammonds:
I've update the code here based on the review. I also update the ap func to take iterables for simulation parameters. This is because 3 gaussians may desired in some cases (i.e. the sum of conductive, sodium and potassium currents).

```python
from neurodsp.sim.cycles import sim_ap_cycle

cycle = sim_ap_cycle(1, 100, (.5, .3, .6), (.1, .1, .1), (0, -1, 1), (-5, 1, 2.5))

plt.plot(cycle)
```
example_double_skew

ryanhammonds:
@TomDonoghue I made a final round of updates. There have been several changes since the last time you reviewed, but it should be ready to merge. Let me know if you'd like to look over the changes, otherwise I'll merge.

TomDonoghue (approved):

PR #260, opened 2021-04-22
[ENH] - Extend access to options for saving figures

TomDonoghue:
I realized the `savefig` implementation that's in LISC seemed to usefully support a couple extra options in the `savefig`, that seemed quite useful. This extends the decorator in NDSP, to allow for setting to save out and close the plot (useful in scripts or loops), and also to allow for passing through kwargs.


ryanhammonds (commented):
Looks good!

PR #263, opened 2021-06-09
[WIP] Phase consistency measures

ryanhammonds:
I was exploring the phase relationship between spikes and lfp and came across a few methods that may be useful additions. Currently, this adds pairwise phase consistency ([Vinck et al. 2010](https://www.sciencedirect.com/science/article/pii/S1053811910000959)). This PR or future PRs could also include phase-locking value and spectral coherence.

### DIFF
--- a/neurodsp/timefrequency/consistency.py
+++ b/neurodsp/timefrequency/consistency.py
@@ -1,0 +1,54 @@
+"""Phase consistency measures."""
+
+from itertools import combinations
+
+import numpy as np
+
+###################################################################################################
+###################################################################################################
+
+def pairwise_phase_consistency(phases):
+    """Compute pairwise phase consistency.
+
+    Parameters
+    ----------
+    phases : 2d array
+        Phases from a wavelet or Fourier analysis, in radians.
+
+    Returns
+    -------
+    avg_distance : float
+        Average pairwise circular distance index.
+    distances : 2d array
+        Pairwise circular distance indices.
+
+    Notes
+    -----
+
+    - distance == -1: inverse phases (i.e. np.pi vs -np.pi)
+    - distance ==  0: pi / 2 phase difference (i.e. np.pi vs np.pi / 2)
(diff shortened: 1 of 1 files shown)


ryanhammonds:
The algorithm should be the same as in the paper but with one exception. When computing the absolute angular distance, the paper reports in Eq. 9 that the absolute angular distance is computed as `d = abs(phi - omega) % np.pi`. When the angle between two phases should be pi, but the reported equation give 0. Updating to `d = abs(phi - omega) % (2 * np.pi)` seems to be more appropriate.

```
phi0 = 0
phi1 = np.pi

d = np.abs(phi0 - phi1) % (2 * np.pi) # returns np.pi
index = (np.pi - 2 * x) / np.pi       # returns -1
```

TomDonoghue:
I think adding these phase measures makes sense! 

In terms of scope, mostly we've stuck to measures applied to single signals (univariate), but I think adding some relatively simple bivariate phase measure things that one might want to use, is totally reasonable!

I have a quick implementation of PLV that I use sometimes that I think should be fairly solid, if you want to use / add that:
```
def phase_locking_value(theta1, theta2):
    """Compute the phase locking value between two signals.
    From: https://dsp.stackexchange.com/questions/25165/phase-locking-value-phase-synchronization
    """

    complex_phase_diff = np.exp(np.complex(0, 1) * (theta1 - theta2))
    plv = np.abs(np.sum(complex_phase_diff)) / len(theta1)

    return plv
```

`theta1` & `theta2` should be 1d arrays of instantaneous phase, I think.

ryanhammonds:
Sorry for the delayed response, I was out last week. Glad you think this was a reasonable addition. I was fascinated by the paper, wrote the code to do it and figured it could be useful to others.

Thanks for the PLV snippet. I'll add that sometime soon. This and #264 can go in after the 2.2.0 release. The new release is ready on my end, but I can always re-review or ask someone else in the lab to help with reviews as I know you're busy, no worries either way!

PR #264, opened 2021-06-15
[MRG] Add multidim decorator to filter_signal

ryanhammonds:
This allows 2d signals to be passed to filter_signal. My use case was filtering a 2d array of signals from a MEA.

I also updated np.stack to np.array in the decorator since it's equivalent in this case and 10x faster.
```python
%%timeit
out = [i+j for i in range(1000) for j in range(1000)]
out = np.stack(out)
```
> 1.27 s ± 10.6 ms per loop (mean ± std. dev. of 7 runs, 1 loop each)

```python
%%timeit
out = [i+j for i in range(1000) for j in range(1000)]
out = np.array(out)
```
> 98.9 ms ± 304 µs per loop (mean ± std. dev. of 7 runs, 10 loops each)

### DIFF
--- a/neurodsp/filt/filter.py
+++ b/neurodsp/filt/filter.py
@@ -2,12 +2,14 @@
 
 from warnings import warn
 
+from neurodsp.utils.decorators import multidim
 from neurodsp.filt.fir import filter_signal_fir
 from neurodsp.filt.iir import filter_signal_iir
 
 ###################################################################################################
 ###################################################################################################
 
+@multidim()
 def filter_signal(sig, fs, pass_type, f_range, filter_type='fir',
                   n_cycles=3, n_seconds=None, remove_edges=True, butterworth_order=None,
                   print_transitions=False, plot_properties=False, return_filter=False):
--- a/neurodsp/filt/filter.py
+++ b/neurodsp/filt/filter.py
@@ -2,14 +2,12 @@
 
 from warnings import warn
 
-from neurodsp.utils.decorators import multidim
 from neurodsp.filt.fir import filter_signal_fir
 from neurodsp.filt.iir import filter_signal_iir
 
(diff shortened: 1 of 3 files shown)


TomDonoghue:
This definitely falls under the category of "poorly documented behaviour", but I think the reason that the multidim decorator wasn't added to the filter functions is that these functions actually do already work on 2d signals. 

On main, for me, this currently works:
```
data_2 = np.ones([10, 1000])
out2 = filter_signal(data_2, 1000, 'bandpass', (8, 12))
```

The most relevant step for this is that the `apply_*` filter functions need to work on 2d arrays, and as set up I think they do, based on how they work with numpy arrays. In this case, not having the `multidim` decorator is preferable, if we can avoid it, since multidim would call the function for each row, redundantly regenerating the filter each time, whereas as long as the apply function can apply the filter across all rows, then everything else only needs to be executed once. 

@ryanhammonds - were you running into some error with applying the filter to 2d arrays? If not, and you can have a quick check that the current approach does generalize beyond 1d, then I think what we should do here is update the docs, rather than add the decorator. 

Also: woah on the different between `stack` and `array`. That's wild. Good catch! We should update that.

ryanhammonds:
@TomDonoghue Thanks for this! You're totally correct, it works for 1d or 2d (but not 3d) signals. I don't see a huge use case for 3d arrays anyways. And like you mentioned it's redundant. I removed the multidim decorator, but did add a note in the docstring that 2d arrays are supported. I also added a small test to show it's does indeed work with 2d arrays as intended.

I was surprised by the the performance diff of stack vs array as well. I _think_ in most cases np.array + list comprehension is typically faster than stack/vstack.

TomDonoghue (approved):

PR #269, opened 2021-07-29
[ENH] Plot spectra accepts 2d arrays

ryanhammonds:
Before, `plot_power_spectra` only accepted 1d array or list of 1d array for freqs and powers. This adds support for 2d arrays. The requirement  to call `.tolist()` on a 2d power array for plotting to work felt strange - espcially since compute_spectrum returns 2d arrays for 2d inputs.

```python
import numpy as np

n_seconds = 10
fs = 1000
freqs = [10, 20]

# Simulate 2d array
sigs = np.zeros((2, n_seconds * fs))

for ind, freq in enumerate(freqs):
    sigs[ind] = sim_oscillation(n_seconds, fs, freq)

freqs, powers = compute_spectrum(sigs, fs)

# Freqs: list of 1d array
plot_power_spectra([freqs, freqs], powers)

# Freqs: 2d array
plot_power_spectra(np.array([freqs, freqs]), powers)

# Powers: list of 1d array
plot_power_spectra(freqs, [powers[0], powers[1]])

# Powers: 2d array
```

### DIFF
--- a/neurodsp/plts/spectral.py
+++ b/neurodsp/plts/spectral.py
@@ -18,9 +18,9 @@ def plot_power_spectra(freqs, powers, labels=None, colors=None, ax=None, **kwarg
 
     Parameters
     ----------
-    freqs : 1d or 2d array
+    freqs : 1d or 2d array or list of 1d array
         Frequency vector.
-    powers : 1d or 2d array
+    powers : 1d or 2d array or list of 1d array
         Power values.
     labels : str or list of str, optional
         Labels for each time series.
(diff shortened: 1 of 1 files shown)


neurodsp/plts/spectral.py
```diff
@@ -18,9 +18,9 @@ def plot_power_spectra(freqs, powers, labels=None, colors=None, ax=None, **kwarg
 
     Parameters
     ----------
-    freqs : 1d array or list of 1d array
+    freqs : 1d or 2d array
         Frequency vector.
-    powers : 1d array or list of 1d array
+    powers : 1d or 2d array
```
TomDonoghue, line 23: Since a list of 1d array inputs still works right, I think we want to keep this listed in the docs, so I think it should probably say something like:
`powers : 1d or 2d array or list of 1d array`

Same for freqs above, I think.

TomDonoghue:
This looks like a super sensible and nice update! I left one small comment on the docs, otherwise, LGTM!

PR #270, opened 2021-08-20
[ENH] Allow min_n_cycles to be zero

ryanhammonds:
This is a small update that allow `min_n_cycles = 0` to be passed to detect_bursts_dual_threshold -  in case one doesn't want to remove detected bursts based on duration. Before this would raise an error.

### DIFF
--- a/neurodsp/burst/dualthresh.py
+++ b/neurodsp/burst/dualthresh.py
@@ -79,7 +79,7 @@ def detect_bursts_dual_threshold(sig, fs, dual_thresh, f_range=None,
 
     # Remove bursts detected that are too short
     # Use a number of cycles defined on the frequency range, if available
-    if f_range and min_n_cycles:
+    if f_range is not None and min_n_cycles is not None:
         min_burst_samples = int(np.ceil(min_n_cycles * fs / f_range[0]))
     # Otherwise, make sure minimum duration is set, and use that
     else:


TomDonoghue:
I think this generally makes sense to allow this, if one really just wants to get the un-tuned detected segments. 

Note, I think that in the existing code, this would actually do the same thing:
```
# Get bursts, with no duration thresholds (note: doesn't matter if f_range is given for this)
detect_bursts_dual_threshold(out, 1000, (1, 10), f_range=None, min_n_cycles=None, min_burst_duration=0)
```
^By forcing it ignore `min_n_cycles`, and use `min_burst_duration`, which can be set to zero, this ends up with `min_burst_samples` of 0, and thus no duration threshold. The quirk is that based on how we currently do the value checks (lines 82-89), managing the possibilities of what gets used is slightly funky, including the error if `min_n_cycles=0` instead of None. 

(Sry for long response lol, this was basically my logic-ing things out as I went through this). 

tl-dr: yeh, I think this makes sense, and allows for this seemingly sensible setup, which currently fails:
```
# Get bursts, setting no duration threshold with `min_n_cycles`
detect_bursts_dual_threshold(out, 1000, (1, 10), f_range=(5, 15), min_n_cycles=0)
```
I don't think I see any undesirable side-effects, or any reason not to allow this, so yeh, might as well!

TomDonoghue (approved):

PR #272, opened 2021-08-24
[ENH] Damped Oscillations

ryanhammonds:
This adds a simulation for the damped oscillations [paper](https://www.biorxiv.org/content/10.1101/2021.08.20.457060v2.full).

```python
import numpy as np
from neurodsp.sim import sim_damped_oscillation

n_seconds = 2
fs = 1000

sig = sim_damped_oscillation(n_seconds, fs, 10, .1)
plot_time_series(times, sig)

sig = sim_damped_oscillation(n_seconds, fs, 10, 1)
plot_time_series(times, sig)

sig = sim_damped_oscillation(n_seconds, fs, 10, 10)
plot_time_series(times, sig)

sig = sim_damped_oscillation(n_seconds, fs, 10, 50)
plot_time_series(times, sig)
```
Simulated Results:
ndsp_sim
Paper Reference:
paper_sim

### DIFF
--- a/neurodsp/sim/periodic.py
+++ b/neurodsp/sim/periodic.py
@@ -276,6 +276,53 @@ def sim_variable_oscillation(n_seconds, fs, freqs, cycle='sine', phase=0, **cycl
     return sig
 
 
+def sim_damped_oscillation(n_seconds, fs, freq, gamma, growth=None):
+    """Simulate a damped relaxation oscillation.
+
+    Parameters
+    ----------
+    n_seconds : float or None
+        Simulation time, in seconds.
+    fs : float
+        Signal sampling rate, in Hz.
+    freq : float
+        Oscillation frequency, in Hz.
+    gamma : float
+        Parametric dampening coefficient.
+    growth : optional, default: None
+        Logistic growth rate to smooth the heaviside step function. If None,
+        a non-smoothed heaviside is used.
+
+    Returns
+    -------
+    sig : 1d array
+        Simulated damped relaxation oscillation.
+
+    References
+    ----------
+    .. [1] Evertz, R., Hicks, D. G., & Liley, D. T. J. (2021). Alpha blocking and 1/fβ spectral
+           scaling in resting EEG can be accounted for by a sum of damped alpha band oscillatory
+           processes. bioRxiv 2021.08.20.457060; DOI: https://doi.org/10.1101/2021.08.20.457060
+
+    Examples
+    --------
+    >>> sig = sim_dampened_oscillation(1, 1000, 10, .1)
+    """
+
+    times = np.arange(0, n_seconds, 1/fs)
+
+    exp = np.exp(-1 * gamma * times)
+    cos = np.cos(2 * np.pi * freq * times)
+
+    if growth is None:
+        logit = 1
+    else:
+        # Smooth heaviside as a logit
+        logit = 1 / (1 + np.exp(-2 * growth * times))
+
+    return exp * cos * logit
+
+
 def make_bursts(n_seconds, fs, is_oscillating, cycle):
     """Create a bursting time series by tiling when oscillations occur.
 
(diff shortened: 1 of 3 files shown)


voytek:
This looks like a nice, simple implementation! Great work.

neurodsp/sim/periodic.py
```diff
@@ -276,6 +276,53 @@ def sim_variable_oscillation(n_seconds, fs, freqs, cycle='sine', phase=0, **cycl
...
+    .. [1] Evertz, R., Hicks, D. G., & Liley, D. T. J. (2021). Alpha blocking and 1/fβ spectral
+           scaling in resting EEG can be accounted for by a sum of damped alpha band oscillatory
+           processes. bioRxiv 2021.08.20.457060; DOI: https://doi.org/10.1101/2021.08.20.457060
+
+    Examples
+    --------
+    >>> sig = sim_dampened_oscillation(1, 1000, 10, .1)
+    """
+
+    times = np.arange(0, n_seconds, 1/fs)
```
TomDonoghue, line 312: Can this use `create_times` instead? Doing so ensures the same procedure for all sims so that they end up exactly the same length and can be summed together.

neurodsp/sim/periodic.py
```diff
@@ -276,6 +276,53 @@ def sim_variable_oscillation(n_seconds, fs, freqs, cycle='sine', phase=0, **cycl
     return sig
 
 
+def sim_damped_oscillation(n_seconds, fs, freq, gamma, growth=None):
+    """Simulate a damped relaxation oscillation.
+
+    Parameters
+    ----------
+    n_seconds : float or None
```
TomDonoghue, line 284: I think None is not an option for this one, right?

neurodsp/sim/periodic.py
```diff
@@ -276,6 +276,53 @@ def sim_variable_oscillation(n_seconds, fs, freqs, cycle='sine', phase=0, **cycl
...
+        Simulation time, in seconds.
+    fs : float
+        Signal sampling rate, in Hz.
+    freq : float
+        Oscillation frequency, in Hz.
+    gamma : float
+        Parametric dampening coefficient.
+    growth : optional, default: None
```
TomDonoghue, line 292: Missing type. float, I think?

neurodsp/tests/sim/test_periodic.py
```diff
@@ -88,6 +88,15 @@ def test_sim_variable_oscillation():
         sig4 = sim_variable_oscillation(None, FS, freqs, cycle='asine', rdsym=rdsyms[1:])
 
 
+def test_sim_damped_oscillation():
+
+    sig1 = sim_damped_oscillation(N_SECONDS, FS, FREQ1, .1)
+    sig2 = sim_damped_oscillation(N_SECONDS, FS, FREQ1, 50)
+
+    assert sig1.sum() < sig2.sum()
+    assert len(sig1) == len(sig2) == int(N_SECONDS * FS)
```
TomDonoghue, line 97: the `check_sig_output` helper function can be used on these, I think  (since they use the standard N_SECONDS & FS), acting as a shortcut to check length / type / nan-ness.

neurodsp/tests/sim/test_periodic.py
```diff
@@ -88,6 +88,15 @@ def test_sim_variable_oscillation():
         sig4 = sim_variable_oscillation(None, FS, freqs, cycle='asine', rdsym=rdsyms[1:])
 
 
+def test_sim_damped_oscillation():
+
+    sig1 = sim_damped_oscillation(N_SECONDS, FS, FREQ1, .1)
+    sig2 = sim_damped_oscillation(N_SECONDS, FS, FREQ1, 50)
+
+    assert sig1.sum() < sig2.sum()
```
TomDonoghue, line 96: It took me a second the understand what this is testing - I think it would be worth a comment, something like: "check that signal with more dampening has lesser overall amplitude"

Also - am I reading the figure in the PR wrong... I think I would expect the sig2, with much higher damping (blue) to have a lesser sum than the lightly damped (yellow) one? Maybe I don't understand this test.

TomDonoghue:
This looks awesome - super clean & quick and nice implementation @ryanhammonds!

I left some comments on small things, but broadly everything looks good, and after these small fixes this should be merged in. Just a reminder to also add this function to the API listing when you do!

Issue #271, opened 2021-08-20
[BUG] Sim bursty oscillation bug


ryanhammonds:
```python
import numpy as np
from neurodsp.sim import sim_bursty_oscillation

np.random.seed(1)

n_seconds = 10
fs = 1000

sim_bursty_oscillation(n_seconds, fs, 15)
```

This traces into updates related to `make_bursts`.

TomDonoghue:
I had a quick look into this, since I worked on these burst functions recently enough to mostly remember how it all works. 

The issue is this part of `make_bursts`:
```
for sig_ind, is_osc in zip(range(0, n_samples, n_samples_cycle), is_oscillating):
    if is_osc:
        burst_sig[sig_ind:sig_ind+n_samples_cycle] = cycle
```

Basically, this bug will occur if a cycle has a number of samples (`n_cycles_cycle`) that doesn't divide evenly into the total number of cycles for the signal (`n_cycles`), and if there is a cycle set to be added as the last cycle. This leads to the index error that gets raised, since the slice into burst_sig ends up being too small for the length of `cycle`. 
The reason it's a bit random if/when it occurs is that it's not-deterministic if there is a cycle set for the last possible cycle position, and so this will randomly work if the last cycle is False (in `is_oscillating`). 

Fixed proposed in #273

TomDonoghue:
Fixed in #273

PR #273, opened 2021-08-29
[BUG] - Fix bug in `sim_bursty_oscillation`

TomDonoghue:
Addresses #271 - details of the problem explained there. 

Basically, because the issue here is trying to add a cycle with a length greater than the remainder of the cycle, the proposed fix here is to skip the add if there are not enough samples enough for the cycle. This means whenever this context arises, regardless of the value of `is_oscillating`, the last few samples will be non-oscillating. 

There is an alternative: instead of leaving trailing zeros, we could update to add a truncated cycle with whatever samples are left. I see no strong reason to prefer one approach over the over, so I think the trailing zeros might be easier. 

For the record: the reason the this bug popped up now is we used to concatenate cycles and then trim the whole signal to the expected n_samples, whereas now (somewhere in the refactors) we pre-allocate exactly the length of the signal from the start (this does mean that with this PR what we do in this scenario is now a bit different from what we used to do). 

The new test explicitly tests the situation that would cause the issue (a cycle set to be added at the last cycle position, without enough samples left the add it). Note that this test would fail on the code in `main`. 

Anyways - long story short, unless anyone has strong thoughts on the strategy taken here, this should be a quick fix that is ready to merge.

### DIFF
--- a/neurodsp/sim/periodic.py
+++ b/neurodsp/sim/periodic.py
@@ -348,7 +348,11 @@ def make_bursts(n_seconds, fs, is_oscillating, cycle):
 
     burst_sig = np.zeros([n_samples])
     for sig_ind, is_osc in zip(range(0, n_samples, n_samples_cycle), is_oscillating):
-        if is_osc:
+
+        # If set as an oscillating cycle, add cycle to signal
+        #   The sample check is to check there are enough samples left to add a full cycle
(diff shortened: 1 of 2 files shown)


ryanhammonds (approved):
Thanks for the fix! I reviewed your code and comments and everything makes sense and looks good to me.

TomDonoghue:
@ryanhammonds - awesome, thanks for checking through it!

PR #274, opened 2021-08-29
[BUG] - Fix doctest typo

TomDonoghue:
This fixes a small typo in the doctest, such that it would fail execution.

As a sidenote: one day we might poke at github actions, to get it to run doctests as well.


ryanhammonds (approved):
Whoops, this was my fault. Thanks for the fix!

TomDonoghue:
@ryanhammonds - no worries at all, just something I happened to notice when i ran the full set of tests, etc.

Issue #276, opened 2021-10-07
sim_poisson_pop bug: undesired normalization


mwprestonjr:
sim\aperiodic.py includes a normalization decorator on line 18 which prevents the function sim_poisson_pop from returning a signal with the desired statistics.

TomDonoghue:
This is in progress in #277 - so I'm going to close this here

PR #277, opened 2021-10-08
[FIX] Default mean and variance of poisson pop to 1

ryanhammonds:
Addresses #276. Before, the Poisson pop would default to a mean of 0 and variance of 1.

### DIFF
--- a/neurodsp/sim/aperiodic.py
+++ b/neurodsp/sim/aperiodic.py
@@ -16,7 +16,7 @@ from neurodsp.sim.transients import sim_synaptic_kernel
 ###################################################################################################
 
 @normalize
-def sim_poisson_pop(n_seconds, fs, n_neurons=1000, firing_rate=2):
+def sim_poisson_pop(n_seconds, fs, n_neurons=1000, firing_rate=2, mean=1, variance=1):
     """Simulate a Poisson population.
 
     Parameters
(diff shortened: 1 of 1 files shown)


mwprestonjr:
Oh okay I understand. I think it would be best if by default the returned signal is Poisson-distributed. Defaulting the mean and variance parameter to None would be fine in this regard. 

However, in general, normalizing the output in this way does not ensure that the signal is non-negative. If we want to make the function more flexible, another option would be to include arguments for the mean and variance of the random distribution from which the signal is sampled.

ryanhammonds:
I see now. I went ahead and removed the decorator since that could result in negative values, and added a `lam` param that directly controls the mean and variance.

mwprestonjr:
There is a related issue using this decorator with sim_random_walk(). This functions takes the arguments `mu' and 'sigma', which are not passed to the decorator (which is looking for 'mean' and 'variance'). Normalization may not be needed here either.

ryanhammonds:
I removed the normalize decorator from sim_random_walk, but I'm not sure what is best here either. It seems like there is bias in simulating an OU process, causing mean and std to be inaccurate, even when the decorator is removed.

I added a norm arg that normalizes to the requested mean/std but it can be set to false in case the estimation bias is of interest. This is a deviation from the api of the other sim funcs so maybe it's not the best decision.

ryanhammonds:
@mwprestonjr 858bf1b contains updates to docs and variance calculation. I redefined variance based on your equation (nice job on figuring that out!) and made a note differentiating sigma as "Wiener scaling", which differs from standard deviation.

TomDonoghue (approved):

TomDonoghue:
Better late than never, I had another quick check through here! Only changes are merging main to check for conflicts and some small aesthetic things, I think this is good to go in!

Thanks @ryanhammonds & @mwprestonjr for digging into this and figuring this out!

PR #280, opened 2021-10-14
[MNT] - Lint clean ups

TomDonoghue:
This is a quick sweep for some lints & clean ups (aesthetic only).

### DIFF
--- a/neurodsp/sim/combined.py
+++ b/neurodsp/sim/combined.py
@@ -92,7 +92,7 @@ def sim_peak_oscillation(sig_ap, fs, freq, bw, height):
     """Simulate a signal with an aperiodic component and a specific oscillation peak.
 
     Parameters
-    -----------
+    ----------
     sig_ap : 1d array
         The timeseries of the aperiodic component.
     fs : float
--- a/neurodsp/spectral/power.py
+++ b/neurodsp/spectral/power.py
@@ -24,7 +24,7 @@ def compute_spectrum(sig, fs, method='welch', avg_type='mean', **kwargs):
     """Compute the power spectral density of a time series.
 
     Parameters
-    -----------
+    ----------
     sig : 1d or 2d array
         Time series.
     fs : float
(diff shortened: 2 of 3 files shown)


neurodsp/sim/aperiodic.py
```diff
@@ -119,64 +119,61 @@ def sim_synaptic_current(n_seconds, fs, n_neurons=1000, firing_rate=2.,
...
         Power law exponent added to chi1 after the knee.
     knee: float
-        Location of the knee in hz.
+        Location of the knee in Hz.
 
     Returns
     -------
-    sig: 1d array
-        Time series with desired power spectrum.
+    sig : 1d array
```
ryanhammonds, line 139: If white space is preferred after the return/arg, it could also be added to the parameters above.

neurodsp/sim/combined.py
```diff
@@ -127,38 +127,35 @@ def sim_peak_oscillation(sig_ap, fs, freq, bw, height):
...
+        (-np.real(sig_ap_hat[ell]) + np.sqrt(np.real(sig_ap_hat[ell]) ** 2 + \
+        (10 ** rel_heights[ell] - 1) * np.abs(sig_ap_hat[ell]) ** 2)) / cosine_norms[ell] \
+        for ell in range(cosine_norms.shape[0])])
 
     # Add cosines with the respective coefficients and with a random phase shift for each one
-    sig_periodic = np.sum(
-        np.array([cosine_coeffs[ell]*np.cos(2*np.pi*freqs[ell]*times + 2*np.pi*np.random.rand())
-                  for ell in range(cosine_norms.shape[0])]),
-        axis=0
-    )
```
ryanhammonds, line 161: Throughout this func, I prefer the old format a bit more. It feels slightly more readable but nbd either way since it's equivalent.

ryanhammonds (commented):
Looks good!

PR #281, opened 2021-10-14
[DOC] - Update API list

TomDonoghue:
This updates to API list to include some new functions that have recently been added. 

It also:
- reorganizes the API list for the `aperiodic` sub-module
- adds an explicit listing of relevant utility functions

### DIFF
--- a/doc/api.rst
+++ b/doc/api.rst
@@ -27,7 +27,6 @@ General Filter Function
 
     filter_signal
 
-
 FIR Filters
 ~~~~~~~~~~~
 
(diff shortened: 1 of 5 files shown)


ryanhammonds (approved):
This all looks good to me! I doubled checked things by building the site and checking the API.

I scripted a check to see which funcs are in the API but aren't imported in an `__init__.py`. For example:
 ```
 from neurdsp.sim.transients import sim_action_potential
```
rather than
```
 from neurdsp.sim import sim_action_potential
```

I figured if it was important enough to be in the API, than it might as well be imported in an init. But feel free to modify the commit I'm about to push.

PR #283, opened 2021-10-15
[MNT] - Tests updates

TomDonoghue:
This PR does a sweep of the tests, with some minor updates, including:
- mildly editing / adding / extending some tests
- fixing the `sim_asine_cycle` func for half cycles (that failed once tests were extended)

Also includes (other things done along the way):
- updating functions to use param checker functions 
- mild aesthetic updates of documentation

### DIFF
--- a/neurodsp/aperiodic/dfa.py
+++ b/neurodsp/aperiodic/dfa.py
@@ -3,6 +3,7 @@
 import numpy as np
 
 from neurodsp.utils.data import split_signal
+from neurodsp.utils.checks import check_param_options
 
 ###################################################################################################
 ###################################################################################################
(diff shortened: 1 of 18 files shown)


ryanhammonds (approved):
Everything looks good here. `check_param_options` reduces a ton of duplicate code.

For `sim_asine_cycle`, I'm not sure about the best fix, but what you have works.

An alternative could be to simulate two separate sine and asine arrays, along with two weight arrays. This may not be ideal either and have other problems though.
```python
# Odd
weights_left = np.array([1, 1, .5, 0, 0])
weights_right = np.array([0, 0, .5, 1, 1])
sig = (sig_left * weights_left ) + (sig_right * weights_right)

# Even
weights_left = np.array([1, 1, 0, 0)
weights_right = np.array([0, 0, 1, 1])
sig = (sig_left * weights_left ) + (sig_right * weights_right)
```

PR #284, opened 2021-11-20
[FIX] Sim knee update

ryanhammonds:
I have been exploring different ways to simulate timescales and remembered we have a `sim_knee` function. After re-exploring this func, I noticed the knee frequency parameter passed into the function does not translate to the simulated knee location. I've update the coefficient equation so the knee is in the proper location.

Below is an example pre/post this PR, simulating a knee at 100hz with a exponent of 2:
sim_knee

Converting the model exponent and knee value to knee frequency, the knee frequency goes from ~10Hz to ~100Hz. However, the exponent becomes less accurate (2.00 -> 1.96).

### DIFF
--- a/neurodsp/sim/aperiodic.py
+++ b/neurodsp/sim/aperiodic.py
@@ -174,7 +174,7 @@ def sim_knee(n_seconds, fs, chi1, chi2, knee):
 
     # Map the frequencies under the (square root) Lorentzian
     #   This will give us the amplitude coefficients for the sinusoids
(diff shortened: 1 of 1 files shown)


TomDonoghue:
Yes - this is a good catch! The current value is the knee "value", not the knee frequency, as the documentation says. I think it makes more sense for the function to take the knee frequency. 

But also, I slightly messed up - and forgot this PR was open before opening #290, with these two PRs now having overlapping changes. Perhaps easiest, since this change is small in terms of code - do you want to redo this commit in #290, and then we'll finish that there? I think after that we'll just need to adjust the test, and then we should be good on this.

ryanhammonds:
Sounds good! I will migrate this over to #290.

PR #285, opened 2022-01-11
[ENH] - Add conversion functions for aperiodic measures

TomDonoghue:
This PR add conversion functions between different related aperiodic measures.

### DIFF
--- a/neurodsp/aperiodic/conversions.py
+++ b/neurodsp/aperiodic/conversions.py
@@ -8,7 +8,7 @@ def convert_exponent_alpha(exponent):
 
     Parameters
     ----------
-    exp : float
+    exponent : float
         Exponent value for a 1/f distribution.
 
     Returns
--- a/neurodsp/aperiodic/conversions.py
+++ b/neurodsp/aperiodic/conversions.py
@@ -11,12 +11,12 @@ def convert_exponent_alpha(exponent):
     Parameters
     ----------
     exponent : float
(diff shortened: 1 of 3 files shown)


neurodsp/aperiodic/conversions.py
```diff
@@ -0,0 +1,255 @@
...
+           Hoyos-Reyes, L. F. (2013). Fractal dimension algorithms and their application
+           to time series associated with natural phenomena. Journal of Physics: Conference
+           Series, 475, 012002. https://doi.org/10.1088/1742-6596/475/1/012002
+
+    Examples
+    --------
+    Convert exponent value to the expected Higuchi fractal dimension:
+
+    >>> convert_exponent_hfd(-1)
+    2.0
```
ryanhammonds, line 204: Under the notes header, it says if exponent = 0, hfd = 2. And then in this example when exponent = -1, hfd = 2.

ryanhammonds (commented):
Looks good!

neurodsp/aperiodic/conversions.py
ryanhammonds: Oops, maybe both are correct. So 0 hfd == -1 hfd == 0 exponent?

TomDonoghue:
Since I don't think there's anything more to do about the quirks of Fractal Dimension conversion here, I'm going to merge this in now - but feel free to tag anything else if you to follow up on anything here!

PR #288, opened 2022-01-26
[MNT] - Refactor spectral rotation

TomDonoghue:
This is a small refactor of the code for doing spectral rotations. 

The reason is that the old organization applied spectral rotations to timeseries to create powerlaw signals, but wasn't organized to allow the user access to this particular computation (as it was buried in `_create_powerlaw`). 

This update moves some code around a bit, to create the new, user-facing, `rotate_timeseries` function, which makes this procedure available to the user to pass in any time series they want. The `_create_powerlaw` now uses this new function.

Open Qs:
- Naming: now that we have `rotate_timeseries`, it might make sense to rename `rotate_powerlaw` -> `rotate_spectrum`
  - This would be a breaking change though, so maybe we want to keep it how it is?
- Organization: it's not entirely clear `spectral/utils` is the right place for this stuff
  - It's also not clear where is better though. Maybe something like `aperiodic/utils/` or `sim/utils`?

Decisions: renamed to `rotate_timeseries`, and moved these to `sim/utils`, but aliased the old function name, for backwards compatibility.


ryanhammonds (approved):
Everything looks good here! @QuirinevEngen was looking for a function to do this.

TomDonoghue:
Awesome, thanks for the check here @ryanhammonds 

I thought your suggestions here make sense - I have rename the `rotate_powerlaw` function, and moved the rotate functions to `sim/utils`, but I've added an alias to the old function name in `spectral/utils`, for backwards compatibility. 

I think this is done now, so I'm merging in!

Issue #287, opened 2022-01-26
Tests failing


TomDonoghue:
On new PRs, our tests are currently failing, on the `test_phase_by_time` function (which has not changed). 

See, for example, here: https://github.com/neurodsp-tools/neurodsp/runs/4945344164?check_suite_focus=true

I don't know why this is happening. It seems to be some newer update, since 3.6 & 3.7 pass, but 3.8 & 3.9 fail. This test checks two vectors for equivalence, and based on the printout on Github actions, they are the same within the range of whats printed. I'm having trouble replicating this error - local environments with what should the same Python / numpy / scipy don't recreate the issue, and it's not obvious how to debug through Github actions. I can't any clear indication of what might have changed in Python / numpy / scipy to create this. 

@ryanhammonds - do you have any idea what might be going on here, and/or any ideas of best strategy to try and figure it out?

ryanhammonds:
This is strange. I'm unable to replicate the issue locally either with 3.8. I also updated my scipy, numpy, pytest versions to the same that are used in GH actions and tests pass. It also looks like the operating system that GH actions uses hasn't changed, nor have the arrays since a random seed it being set in the pytest config.

I'll look into this some more and see if I can figure out what's going on.

TomDonoghue:
Fixed by #289 - thanks Ryan!

PR #289, opened 2022-01-26
[MRG] Fix Hilbert Test

ryanhammonds:
I'm using this PR to trigger GH actions and to figure out why 3.8/3.9 tests are failing.

### DIFF
--- a/neurodsp/tests/timefrequency/test_hilbert.py
+++ b/neurodsp/tests/timefrequency/test_hilbert.py
@@ -57,6 +57,10 @@ def test_phase_by_time(tsig, tsig_sine):
     expected_answer = np.array(\
         [time-np.pi/2 if time <= 3*np.pi/2 else time-5*np.pi/2 for time in times])
 
+    # Round to the sixth decimal place and convert from (-pi, pi) to (0, 2pi)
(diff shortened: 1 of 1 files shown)


ryanhammonds:
@TomDonoghue, this is a fix for #287. I wasn't able to reproduce the error locally. I also installed docker to reproduced the gh action environment, and the test would still pass. I had to debug the issue with print statements + triggering gh actions through this PR :cry:.

Anyways, what was happening was the was one pair of values, pi and -pi, that should have been equal. I ended up rounding to 6 decimals places and then using modulo 2pi such that pi == -pi.

TomDonoghue (approved):

TomDonoghue (approved):

TomDonoghue:
@ryanhammonds - thanks a ton for doing the deep dive on this! What an annoying one - it's wild that nothing seems to recreate this error other than one specific Python versions on Github...

It _sorta_ makes sense that it's a failing asserting by np.pi & -np.pi (in the sense that these are effectively equivalent but not literally the same values) - but the quirks that made this pop up, in this way, are still totally baffling. 

Anyways, great work chasing it out, and thanks for the work on this one! I'm going to merge in now!

Issue #291, opened 2022-01-28
[MNT] - Make times vector optional in timeseries plots?


TomDonoghue:
I sometimes find it a bit annoying to not be able to use `plot_timeseries` with just passing a signal, with no time vector (if, for example, I'm playing around with sims, and don't want to have to bother creating the times vector). 

Making the plot work with just a signal (no times) would be pretty easy - just make times and sigs default to None, and then at the top of the function, check it either are none, assume the other is a signal, and if so, set to plot and label the timeseries based on samples, rather than time (since we won't know the time definition). 

This is a quick / easy update for the sake of laziness, but is a little messy - it basically creates two call signatures for the function, including passing a single vector as first positional argument, that gets interpreted as a a signal. A potentially useful aspect is not only supporting this use case, but extending the function to plotting signals by sample. 

So, thoughts: is this a worthy update, or is the mess not worth it? If it seems worth trying, I can make a quick PR. 

cc: @ryanhammonds

ryanhammonds:
I'm in favor of this! I use `plt.plot` a ton to avoid needing to specify a time array in a hurry. Maybe something like: `plot_timeseries(None, sig)`?

TomDonoghue:
done in #295

PR #295, opened 2022-01-29
[MNT] - Update plot_time_series to make times optional

TomDonoghue:
Responds to #291 

This update means one doesn't have to pass in a times vector to `plot_time_series`, and if not provided it will be plotted in terms of samples instead. 

@ryanhammonds - I like your idea of simply supporting `plot_time_series(None, sig)`, which is what this does. 

My initial idea had be to support `plot_time_series(sig)`, in which case the function would check that the second input is empty, infer the first input is the signal and proceed from there. It would be easy to add this, but this is where it makes the API a little messy. Do we want to add this additional shortcut, or should we keep it like this?

Note that due to how they wrap `plot_time_series`, this update makes this option available to `plot_instantaneous_measure` and `plot_bursts` as well.

### DIFF
--- a/neurodsp/plts/time_series.py
+++ b/neurodsp/plts/time_series.py
@@ -8,6 +8,7 @@ import matplotlib.pyplot as plt
 
 from neurodsp.plts.style import style_plot
 from neurodsp.plts.utils import check_ax, savefig
+from neurodsp.utils.data import create_samples
 from neurodsp.utils.checks import check_param_options
 
 ###################################################################################################
@@ -20,8 +21,9 @@ def plot_time_series(times, sigs, labels=None, colors=None, ax=None, **kwargs):
 
     Parameters
     ----------
-    times : 1d or 2d array, or list of 1d array
+    times : 1d or 2d array, or list of 1d array, or None
         Time definition(s) for the time series to be plotted.
+        If None, time series will be plotted in terms of samples instead of time.
     sigs : 1d or 2d array, or list of 1d array
         Time series to plot.
     labels : list of str, optional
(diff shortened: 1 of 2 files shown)


neurodsp/plts/time_series.py
```diff
@@ -48,9 +50,15 @@ def plot_time_series(times, sigs, labels=None, colors=None, ax=None, **kwargs):
 
     ax = check_ax(ax, (15, 3))
 
-    times = repeat(times) if (isinstance(times, np.ndarray) and times.ndim == 1) else times
     sigs = [sigs] if (isinstance(sigs, np.ndarray) and sigs.ndim == 1) else sigs
 
+    xlabel = 'Time (s)'
+    if times is None:
+        times = create_samples(len(sigs[0]))
+        xlabel = 'Samples'
```
ryanhammonds, line 58: `xlabel` should default to `Time (s)` somewhere. When I plot, specifying a time array, there is no xlabel. Maybe:
```
xlabel = 'Samples' if times is None else 'Time (s)'
times = create_samples(len(sigs[0])) if times is None else times
```
or 
```
if times is None:
    times = create_samples(len(sigs[0]))
    xlabel = 'Samples'
else:
    xlabel = 'Time (s)'
```

ryanhammonds (commented):
Looks good! I only had one comment about defaulting `xlabel`.

I think `plot_time_series(sig)` looks nice but makes the API potentially confusing and would result in errors if additional kwargs are passed. Something like `plot_time_series(sig, color='r')` would raise an error since two positionals are expected.

neurodsp/plts/time_series.py
TomDonoghue: Hmmm...., that's weird that the times label isn't being added - I'm pretty sure that was working when I tested this.

`xlabel` is defined in line 55, which should provide it with a default value that gets used if the if statement in line 56 is not invoked. This is (I think) effectively equivalent to the suggestions you mentioned, so I'm not sure why it wouldn't get applied?

ryanhammonds: Whoops, sorry! I don't know how I didn't see line 55. The plot  I was using had additional custom plt calls that was messing with the xlabel.

TomDonoghue:
Thanks for checking in on this!

Yeh, I don't think making `plot_time_series(sig)` is necessarily worth it. I think something like `plot_time_series(sig, color='r')` could be made to work by declaring `times` and `sig` as default None (hence, optional), but i think it's messy enough to not be worth the small gains - so I'll leave this update as is, and merge this in

PR #296, opened 2022-01-29
[MNT] - Allow for passing figsize to plots

TomDonoghue:
Our plot functions all have default sizes, which could be bypassed by passing in an Axis object. This extends the flexibility for controlling figure size, by allowing for passing in a `figsize` argument directly to each plot argument. 

In general, this `figsize` argument is caught by `kwargs`, and when creating a new axis, the figures will still default to the same size they always have, but this can now be over-written by the `figsize` input.

### DIFF
--- a/neurodsp/plts/filt.py
+++ b/neurodsp/plts/filt.py
@@ -10,7 +10,7 @@ from neurodsp.plts.utils import check_ax, savefig
 ###################################################################################################
 
 @savefig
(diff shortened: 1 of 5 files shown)


ryanhammonds (approved):
Everything looks good here!

TomDonoghue:
awesome, thanks for checking! merging in now

Issue #262, opened 2021-05-06
Potential ERP simulation


TomDonoghue:
This [paper](https://www.nature.com/articles/s41598-018-20423-z) has what looks to be a simple formula for simulating ERPs:

![Screen Shot 2021-05-06 at 12 06 54 PM](https://user-images.githubusercontent.com/7727566/117330346-959cab00-ae63-11eb-9750-3debb75aed66.png)

It's a pretty simple approach (basically, windowing a sine-wave), but seems to work for their purposes, and might be an interesting addition to our `transients` simulations.

TomDonoghue:
This is in progress in #298, so I'm going to close this here now

PR #298, opened 2022-02-01
[WIP/ENH] - Add erp sim func

TomDonoghue:
Responds to #262 - adds a simulation function to create an ERP complex, simulated as a decaying (damped) sine wave. 

@ryanhammonds: could you take a quick look at this - do you think it's something worth adding? It's a very simple approach for adding ERPs, but seems worth adding something like this. If it looks good to you, the main potential update is to update it to vary the start time (I wasn't immediately clear on how to do this).

### DIFF
--- a/doc/api.rst
+++ b/doc/api.rst
@@ -248,8 +248,8 @@ Aperiodic Signals
     sim_frac_gaussian_noise
     sim_frac_brownian_motion
 
-Transients & Cycles
-~~~~~~~~~~~~~~~~~~~
+Transients
+~~~~~~~~~~
 
 .. currentmodule:: neurodsp.sim.transients
 .. autosummary::
@@ -257,6 +257,10 @@ Transients & Cycles
 
     sim_synaptic_kernel
     sim_action_potential
+    sim_damped_erp
+
+Cycles
+~~~~~~
 
 .. currentmodule:: neurodsp.sim.cycles
 .. autosummary::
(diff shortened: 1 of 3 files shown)


ryanhammonds (approved):
This is cool! This kind of reminds me of a damped oscillations, with an extra linear term.

I looked into how to have a start time, but the equation in the paper doesn't seem to work. I guess alternatives could be to np.roll the signal or to redefine the time array after simulating.

TomDonoghue:
@ryanhammonds - yeh, it is pretty similar to a damped oscillator, basically, good point!

I also originally played around with manipulating the start time based on the original equation, and couldn't get it to work but I wasn't sure if I was just messing it up. Using np.roll or something like that seems like a good way to do it - do you want to try adding that?

Otherwise, the other possible thing to check in on is the name (as always...). I think `sim_erp` is obviously relevant, but thinking forward, I somewhat imagine we might add other ways to do this, and don't necessary want this is a the default version - in which case maybe we relabel it something a bit more specific?

ryanhammonds:
@TomDonoghue I added a `time_start` arg that let's one specify where in the signal the erp should start, assuming samples prior to the start time should be zero. I did this over np.roll to avoid samples wrapping to the beginning of the signal in cases where slow decays are used.

I'm not too sure about a good func name. Maybe `sim_damped_erp`?

TomDonoghue:
@ryanhammonds - awesome, thanks for adding the variable start time, that LGTM!

I've changed the name to `sim_damped_erp` - I think it's a good / descriptive label for this particular approach of creating ERPs. I've added that name to the API list, so this should be good to go, and I'm going to merge it in now!

PR #302, opened 2022-04-16
[ENH] - Add spectral measures

TomDonoghue:
These are some helper functions I find myself copying around and using in a few different places - so I figured that it might make sense to add them to NeuroDSP to have a stable / shared implementation. 

Added functions:
- `compute_absolute_power`
- `compute_relative_power`
- `compute_band_ratio`

Note: these are useful functions to compute these measures, especially for comparisons sake, but these are generally not measures I would _recommend_ (we might want to add some kind of disclaimer / note to these functions).

### DIFF
--- a/neurodsp/utils/core.py
+++ b/neurodsp/utils/core.py
@@ -12,7 +12,7 @@ def get_avg_func(avg_type):
 
     Parameters
     ----------
-    avg_type : {'mean', 'median'}
+    avg_type : {'mean', 'median', 'sum'}
         The type of averaging function to use.
 
     Returns
--- a/neurodsp/utils/core.py
+++ b/neurodsp/utils/core.py
@@ -17,15 +17,17 @@ def get_avg_func(avg_type):
 
     Returns
     -------
-    avg_func : callable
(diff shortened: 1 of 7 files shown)


neurodsp/utils/core.py
```diff
@@ -17,15 +17,17 @@ def get_avg_func(avg_type):
...
+    check_param_options(avg_type, 'avg_type', ['mean', 'median', 'sum'])
 
     if avg_type == 'mean':
-        avg_func = np.mean
+        func = np.mean
     elif avg_type == 'median':
-        avg_func = np.median
+        func = np.median
+    elif avg_type == 'sum':
+        func = np.sum
```
ryanhammonds, line 31: sum could be added to the docstring:  `{'mean', 'median', 'sum'}`

ryanhammonds (approved):
Looks good!

neurodsp/utils/core.py
TomDonoghue: Oops, good catch! Fixed!

PR #305, opened 2022-04-21
[MNT] - Update utilities usage in sim/periodic

TomDonoghue:
This PR is some minor clean-ups to using helper utilities in `sim/periodic`. The goal is to more consistently use helper utilities to make sure we are doing the same thing in the same ways. 

Updates:
- update to use `compute_nsamples` more consistently
- add a new helper function to compute the time length of a single cycle
- update to use `compute_cycle_nseconds` in simulation functions

Note for review: this should be a pretty straight forward update, I think, but given it changes some basic underlying estimations when simulating signals, it's probably worth keeping an eye on if this changes any sims in any unexpected ways (I haven't noticed anything so far).

### DIFF
--- a/neurodsp/utils/data.py
+++ b/neurodsp/utils/data.py
@@ -110,6 +110,35 @@ def compute_nseconds(sig, fs):
     return len(sig) / fs
 
 
+def compute_cycle_nseconds(freq, fs=None):
+    """Compute the length, in seconds, for a single cycle at a particular frequency.
+
+    Parameters
+    ----------
+    fs :  float
+        Sampling rate, in Hz.
+    freq : float, optional
+        Oscillation frequency, in Hz.
+        If provided, this is used to get a cycle length optimized for the sampling rate.
+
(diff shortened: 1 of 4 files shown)


neurodsp/utils/data.py
```diff
@@ -110,6 +110,35 @@ def compute_nseconds(sig, fs):
...
 
 
+def compute_cycle_nseconds(freq, fs=None):
+    """Compute the length, in seconds, for a single cycle at a particular frequency.
+
+    Parameters
+    ----------
+    fs :  float
+        Sampling rate, in Hz.
+    freq : float, optional
```
ryanhammonds, line 120: These parameters are mixed up. freq is positional and fs is optional.

ryanhammonds (approved):

neurodsp/utils/data.py
TomDonoghue: Ooops! Good catch - fixed!

PR #307, opened 2022-09-21
[MNT] - Add CI testing against python 3.10

TomDonoghue:
This PR adds testing using Python version 3.10 to our continuous integration. 

Note: when I had a quick look for adding this, I saw some stuff about how writing "3.10" is required to enforce 3.10, otherwise it can end up with 3.1. I didn't check this, but seems to work, so I think this should be good.


ryanhammonds (approved):
The 3.10 build is passing and everything looks good! Thanks for adding this.

PR #314, opened 2023-07-05
[MNT] - Update makefile options for distributions

TomDonoghue:
Adds some helper utilities to the module Makefile to assist with managing / publishing distributions.

Same as https://github.com/fooof-tools/fooof/pull/275
@ryanhammonds - in terms of 'review', just tagging you in to so you see this / to check you're happy to add across modules!


ryanhammonds (approved):
Looks good! This makes memorizing those commands easier.

Issue #316, opened 2023-07-19
Add multitaper method for spectral density estimation


SM-Figueroa:
Hi, [MJ](https://github.com/mwprestonjr) and I have been talking about possibly implementing a multi-taper method for spectral density estimation within neurodsp. I've heard that there is some interest in this among other members of VoytekLab. For the time being, I have opened a pull request with an initial implementation of this method under neurodsp/spectral/power.py. Any notes of advice or collaborators are greatly appreciated. Let me know if there's anything else I can do.

[multi-taper method inspiration](https://www.youtube.com/watch?v=OBwnmiVT9TE)

[pull request](https://github.com/neurodsp-tools/neurodsp/pull/317)

TomDonoghue:
Completed in #317

PR #317, opened 2023-07-19
[ENH] - Add multitaper method for spectral density estimation

SM-Figueroa:
This is an initial attempt at implementing a multi-taper method for estimating power spectral density. Refer to the issue raised [here](https://github.com/neurodsp-tools/neurodsp/issues/316) for more details.

### DIFF
--- a/neurodsp/spectral/power.py
+++ b/neurodsp/spectral/power.py
@@ -66,7 +66,7 @@ def compute_spectrum(sig, fs, method='welch', avg_type='mean', **kwargs):
         return compute_spectrum_medfilt(sig, fs, **kwargs)
 
     elif method == 'multitaper':
-        return compute_spectrum_multitaper(sig, fs **kwargs)
+        return compute_spectrum_multitaper(sig, fs, **kwargs)
 
 
 @multidim(select=[0])
@@ -278,7 +278,7 @@ def compute_spectrum_multitaper(sig, fs, bandwidth, num_windows):
 
     # Compute fourier on signal weighted by each slepian sequence
     freqs = np.fft.rfftfreq(sig_len, 1. /fs)
-    spectrums = np.fft.rfft(slepian_sequences[:, np.newaxis]*sig)
+    spectrums = np.abs(np.fft.rfft(slepian_sequences[:, np.newaxis]*sig))**2
 
     # Average spectrums of each sequence for final result
     spectrum = spectrums.mean(axis=0)
--- a/neurodsp/spectral/power.py
+++ b/neurodsp/spectral/power.py
@@ -299,7 +299,7 @@ def compute_spectrum_multitaper(sig, fs, bandwidth=None, n_tapers=None,
     >>> from neurodsp.sim import sim_combined
     >>> sig = sim_combined(n_seconds=10, fs=500,
     ...                    components={'sim_powerlaw': {}, 'sim_oscillation' : {'freq': 10}})
-    >>> freqs, spec = compute_spectrum_mulitaper(sig, fs=500)
+    >>> freqs, spec = compute_spectrum_multitaper(sig, fs=500)
     """
 
     from scipy.signal.windows import dpss
--- a/neurodsp/spectral/power.py
+++ b/neurodsp/spectral/power.py
@@ -16,7 +16,7 @@ from neurodsp.utils.checks import check_param_options
 from neurodsp.utils.outliers import discard_outliers
 from neurodsp.timefrequency.wavelets import compute_wavelet_transform
 from neurodsp.spectral.utils import trim_spectrum
-from neurodsp.spectral.checks import check_spg_settings
+from neurodsp.spectral.checks import check_spg_settings, check_mt_settings
 
 ###################################################################################################
 ###################################################################################################
@@ -246,7 +246,7 @@ def compute_spectrum_medfilt(sig, fs, filt_len=1., f_range=None):
     return freqs, spectrum
 
 
-def compute_spectrum_multitaper(sig, fs, bandwidth, num_windows):
+def compute_spectrum_multitaper(sig, fs, bandwidth=None, num_windows=None):
     """Compute the power spectral density using the multi-taper method.
 
     Parameters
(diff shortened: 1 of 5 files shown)


neurodsp/spectral/power.py
```diff
@@ -65,6 +65,9 @@ def compute_spectrum(sig, fs, method='welch', avg_type='mean', **kwargs):
     elif method == 'medfilt':
         return compute_spectrum_medfilt(sig, fs, **kwargs)
 
+    elif method == 'multitaper':
+        return compute_spectrum_multitaper(sig, fs **kwargs)
```
ryanhammonds, line 69: missing comma between fs and kwargs

neurodsp/spectral/power.py
```diff
@@ -241,3 +244,47 @@ def compute_spectrum_medfilt(sig, fs, filt_len=1., f_range=None):
         freqs, spectrum = trim_spectrum(freqs, spectrum, f_range)
 
     return freqs, spectrum
+
+
+def compute_spectrum_multitaper(sig, fs, bandwidth, num_windows):
```
ryanhammonds, line 249: this func can be aliased in `spectral/__init__.py` so it can be directly imported from neurodsp.spectral like the other funcs

neurodsp/spectral/power.py
```diff
@@ -241,3 +244,47 @@ def compute_spectrum_medfilt(sig, fs, filt_len=1., f_range=None):
...
+
+    # Compute signal length based on input shape
+    sig_len = sig.shape[sig.ndim - 1]
+
+    # Create slepian sequences
+    slepian_sequences = dpss(sig_len, bandwidth/2, num_windows)
+
+    # Compute fourier on signal weighted by each slepian sequence
+    freqs = np.fft.rfftfreq(sig_len, 1. /fs)
+    spectrums = np.fft.rfft(slepian_sequences[:, np.newaxis]*sig)
```
ryanhammonds, line 281: Does this need abs()**2 to move to psd? I think rfft gives fourier coefficients - some were negative and plotting log power return nans.

ryanhammonds (commented):
Thanks for making this! I left some general feedback/questions. I compared this to the mne's psd_array_multitaper and the spectra here seem noiser. I wonder if mne windows the signal and then compute tapers on a window basis, averages per window, and averages across windows? I'll have to take a look into mne more. Here's a comparison:

Screenshot 2023-07-20 at 11 47 08 AM

neurodsp/spectral/power.py
SM-Figueroa: Done! Should work now.

neurodsp/spectral/power.py
SM-Figueroa: I believe this is correct. After making this change the resulting spectra looks much similar to that of mne's `psd_array_multitaper` (shown below). Default parameters were used for `psd_array_multitaper`, whereas `bandwith = 8 * (sfreq / n_times)` and `num_windows = 7` in `compute_spectrum_multitaper`. I believe this is equivalent to the default values used in `psd_array_multitaper`.

image


image

TomDonoghue:
The tests weren't running for some reason, so I made an edit to add the new func to the API list, to quickstart the tests, and that seemed to work. 

I'm on board with adding a basic multitaper option to ndsp. Since I think @mwprestonjr has a bit more experience with this method, I'm curious on his thoughts on this, and I can also do a quick review if useful. 

@SM-Figueroa - can you also add a test that runs the new multitaper function? You can see examples in the other test files - it doesn't have to be a rigorous check for accuracy (since we probably don't know what the exact numerical outputs should be), but at least include a "smoke test" that runs the function, checks the inputs & outputs are all as expected in shape, type, etc (this at least catches if the function stops running / returning properly).

mwprestonjr:
I'll have some time to review this in the next few days.

mwprestonjr:
I added a check_mt_settings function which sets default parameter values for compute_spectrum_multitaper (analogous to check_spg_settings).  This function also uses the bandwidth parameter to compute NW, which should be used as the second argument to scipy.DPSS. With this change, the results should be comparable to mne.time_frequency.psd_array_multitaper. 

It's worth noting that later tapers in the Slepian sequence have lower spectral concentration. Babadi and Brown (IEEE 2014) suggest limiting the number of tapers to significantly less than n_samples * bandwidth * dt * 0.5 for this reason; MNE weights the tapered spectra based on the concentration ratios, and provides an option (low_bias=True/False) for dropping tapers with ratios less than 0.9: in addition to this weighting method, MATLAB's multitaper method includes an 'adaptive' weighting method that is frequency dependent. In short, there doesn't seem to be a single correct way to weight and recombine tapers. But we should probably have an option for weighting/dropping tapers. I can work on implementing this.

SM-Figueroa:
I added a simple output test comparing the shape of the returned frequency array to the power array as done for other methods in test_power.py. I'm unsure of how to test the actual output values of the method for accuracy. Any thoughts would be much appreciated.

TomDonoghue:
Hey @SM-Figueroa & @mwprestonjr - sorry for dropping off here! 

In terms of status - it seems we have a basic working implementation, and a basic smoke test (checks the function runs), right - with no particular known issues? In terms of testing accuracy - that's actually quite hard to do to any detail. In `test_compute_spectrum_welch` there are some examples that can be adapted - for example, testing that a 10 hz sine wave has a high power value around 10, low power values otherwise, etc, which can be used to test for the function doing approximately the right thing. 

To @mwprestonjr's point about weighting tapers - this seems like an important extension. Not knowing too much about this - do you think we can / should finish and merge this, and then try to extend it later, or is the weighting so important that it should really be added here?

mwprestonjr:
Hey @TomDonoghue! That's right. I think adding a basic test of accuracy as described would be good. 

I went ahead and implemented the weighting procedure. There's now a 'low_bias' argument that's True by default, which drops tapers with low concentration ratios. I also added an 'eigenvalue_weighting' option that's True by default, which weights each spectral estimate by the concentration ratio of its respective taper. 

The procedure and results are similar to MNE. 
image

mwprestonjr (approved):

TomDonoghue (approved):

TomDonoghue:
This all looks great - thanks for the updates @mwprestonjr!

Everything here looks good to me, so I'm going to go ahead and merge this in! Thanks for the work on this!

PR #319, opened 2023-08-29
[ENH] - Plot updates & extensions

TomDonoghue:
Some additions & tweaks to plotting including:
- add a plot for multiple time series (plotted with a vertical offset)
- add a combined 'timeseries & psd' plot
- style tweaks that allow for settings ticks and ticklabels through `plot_style`
- update styling descriptions and implementation, following fooof updates (https://github.com/fooof-tools/fooof/pull/300)
- add tight_layout as a custom plot style, meaning it can be passed in as an argument to turn on/off, and also apply with warning suppression, to stop annoying warnings that the plot layout has changed

The upshot of the style update is that something like this now works:
```
# Plot with a neurodsp plot function with `plot_style`, turning off xticks
plot_function(data, xticks=[])
```

Note: the tests on newer versions of python were failing on a doctest, where an answer that used to come out at 2.0 is now 2.000000000000026 (or something) - so I added rounding to this check.

### DIFF
--- a/neurodsp/plts/combined.py
+++ b/neurodsp/plts/combined.py
@@ -3,6 +3,7 @@
 import matplotlib.pyplot as plt
 
 from neurodsp.spectral import compute_spectrum
+from neurodsp.utils.data import create_times
 from neurodsp.spectral.utils import trim_spectrum
 from neurodsp.plts.spectral import plot_power_spectra
 from neurodsp.plts.time_series import plot_time_series
@@ -12,24 +13,27 @@ from neurodsp.plts.utils import savefig
 ###################################################################################################
 
 @savefig
-def plot_timeseries_and_spectrum(times, sig, fs, f_range=None, spectrum_kwargs=None,
-                                 ts_kwargs=None, psd_kwargs=None, **plt_kwargs):
+def plot_timeseries_and_spectrum(sig, fs, ts_range=None, f_range=None, spectrum_kwargs=None,
+                                 start_val=None, ts_kwargs=None, psd_kwargs=None, **plt_kwargs):
     """Plot a timeseries together with it's associated power spectrum.
 
     Parameters
     ----------
-    times : 1d array, or None
-        Time definition(s) for the time series to be plotted.
-        If None, time series will be plotted in terms of samples instead of time.
-    sigs : 1d array
+    sig : 1d array
         Time series to plot.
     fs : float
         Sampling rate, in Hz.
+    ts_range : list of [float, float], optional
+        The time range to restrict the time series to.
+        For visualization only - the power spectrum is computed over the entire time series.
     f_range : list of [float, float], optional
         The frequency range to restrict the power spectrum to.
     spectrum_kwargs : dict, optional
         Keyword arguments for computing the power spectrum.
         See `compute_spectrum` for details.
+    start_val : float, optional
+        The starting value for the time definition for the time series.
+        If not provided, defaults to zero.
     ts_kwargs : dict, optional
         Keyword arguments for customizing the time series plot.
     psd_kwargs : dict, optional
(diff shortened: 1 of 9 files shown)


neurodsp/plts/combined.py
```diff
@@ -0,0 +1,63 @@
...
+from neurodsp.spectral.utils import trim_spectrum
+from neurodsp.plts.spectral import plot_power_spectra
+from neurodsp.plts.time_series import plot_time_series
+from neurodsp.plts.utils import savefig
+
+###################################################################################################
+###################################################################################################
+
+@savefig
+def plot_timeseries_and_spectrum(times, sig, fs, f_range=None, spectrum_kwargs=None,
```
ryanhammonds, line 15: Could `times` be computed in this function, since `fs` is passed, or do you think it should be an arg? Actually since `plot_time_series` requires it too, this makes sense. But this highlights why I never use plot_time_series and opt for plt.plot(sig) instead, maybe I'm just lazy lol.

neurodsp/plts/combined.py
```diff
@@ -0,0 +1,63 @@
...
+    fs : float
+        Sampling rate, in Hz.
+    f_range : list of [float, float], optional
+        The frequency range to restrict the power spectrum to.
+    **spectrum_kwargs : dict, optional
+        Keyword arguments for computing the power spectrum.
+        See `compute_spectrum` for details.
+    **ts_kwargs : dict, optional
+        Keyword arguments for customizing the time series plot.
+    **psd_kwargs : dict, optional
```
ryanhammonds, line 35: from `**psd_kwargs` up to `**spectrum_kwargs`, these aren't being capture by **kwargs, so I think the **'s can be removed since it's explicitly a dict and not an unpacked dict.

neurodsp/plts/time_series.py
```diff
@@ -158,3 +152,54 @@ def plot_bursts(times, sig, bursting, ax=None, **kwargs):
 
     bursts = np.ma.array(sig, mask=np.invert(bursting))
     plot_time_series(times, [sig, bursts], ax=ax, **kwargs)
+
+
+@savefig
+@style_plot
+def plot_multi_time_series(times, sigs, colors=None, ax=None, **plt_kwargs):
```
ryanhammonds, line 159: Could be added to init for lazier imports.

neurodsp/plts/combined.py
ryanhammonds: Also, the func could be added to init for lazier imports.

ryanhammonds (commented):
Looks good! I added a few optional comments. Something to think about for the future though is a separate plotting package the signals/psds, as the plotting functionality here grows.

neurodsp/plts/combined.py
TomDonoghue: Ooops, yeh, you're totally right - fixed!

neurodsp/plts/combined.py
TomDonoghue: Yeh, that's a good point - this function doesn't totally _need_ `times`, it can recompute from sig length and fs, so I guess the main question is whether there is any reason to try and map the API here to match plot_timeseries (?)

It didn't seem to me there was a strong reason to keep times - so I've pushed an update to drop it, and recompute. I also added a convenience argument to restrict the plotted time series range, and an option to update the start value (since the only real benefit of having `times` is that recomputing assume a time range of 0-siglength, whereas providing a specific times vector allows for anything). Do you think this is better?

In terms of using plt.plot instead of plot_time_series - they are sorta the same in the end hahaha, but (just double-checking you know) you can pass None as times and then you get it plotted by sample, instead of having to manually compute times!

Also - I tried adding to __init__, but the tests failed for a circular import, which I didn't try and chase out. Should we try and address that do you think?

TomDonoghue:
Thanks for checking @ryanhammonds! I made some updates based on your comments, if you want to check the update for the multi plot in particular (since what the API for that _should_ be is not so obvious). 

And yeh, I agree - as things grow, we should definitely consider "what belongs where" and if/when we might consider splitting things out from ndsp. I've been meaning to do a proper sweep of code for a bit, so maybe let's check in on this sometime soon, as we check in on what goes into 2.3, what goes after, and any things we need to figure out on the roadmap otherwise.

neurodsp/plts/combined.py
ryanhammonds: Cool! I like the change. Another optional is to make `times` an optional kwarg. But using `start_val` also works. Up to you, the are the same to me.

I forget / didn't know about passing None to `plot_time_series`. I may use it more now! Especially since it removes the need for styling, etc.

Hmm, yeah that's odd. I spent a couple minutes looking at it and it's not obvious. Seems like something to do with these interactions:
```
compute_spectrum -> filter_signal_fir -> plot_filter_properties
compute_spectrum -> plot_timeseries_and_spectrum
plot_timeseries_and_spectrum and plot_filter_properties in __init__?
```
A quick/lazy solution is to drop the `neurodsp.spectral` imports inside the body of `plot_timeseries_and_spectrum`.

Issue #321, opened 2023-09-13
MNT - Update `filter_signal` to check inputs in relation to filter type


TomDonoghue:
Currently, the `filter_signal` function accepts a `filter_type` input {FIR, IIR}, and several inputs, some of which are specific to choosing a particular filter type. However, it does not check if the inputs are consistent with the filter type, including that the function will default to FIR without the user specifying. 

It would be nice if the function was a bit smarter about this, including:
- checking if the set of inputs are consistent with the value for `filter_type`
- adding a warning / error if the values for filter_type and arguments are inconsistent
- and/or if the filter_type is not otherwise specified, updating type to match the input parameters

TomDonoghue:
This is now addressed in #322 - so I'm going to close it here

PR #322, opened 2023-09-23
[MNT] - Update filter type checks & inference

TomDonoghue:
Responds to #321 

Previously, one could end up with an inconsistent set of input parameters to the general `filter_signal` function - for example, passing in `butterworth_order` without explicitly setting IIR `filter_type`. This update switches `filter_type` to default to None, an if not explicitly set, it is inferred from the given parameters, with additional updates to check that the given set of parameters are consistent.

### DIFF
--- a/neurodsp/filt/filter.py
+++ b/neurodsp/filt/filter.py
@@ -9,7 +9,7 @@ from neurodsp.utils.checks import check_param_options
 ###################################################################################################
 ###################################################################################################
 
-def filter_signal(sig, fs, pass_type, f_range, filter_type='fir',
+def filter_signal(sig, fs, pass_type, f_range, filter_type=None,
                   n_cycles=3, n_seconds=None, remove_edges=True, butterworth_order=None,
                   print_transitions=False, plot_properties=False, return_filter=False):
     """Apply a bandpass, bandstop, highpass, or lowpass filter to a neural signal.
@@ -39,8 +39,8 @@ def filter_signal(sig, fs, pass_type, f_range, filter_type='fir',
         Length of filter, in seconds, if using an FIR filter.
         This parameter overwrites `n_cycles`.
     filter_type : {'fir', 'iir'}, optional
-        Whether to use an FIR or IIR filter.
-        The only IIR filter offered is a butterworth filter.
+        Whether to use an FIR or IIR filter. IIR option is a butterworth filter.
+        If None, type is inferred from input parameters, and/or defaults to FIR.
     remove_edges : bool, optional, default: True
         If True, replace samples within half the kernel length to be np.nan.
         Only used for FIR filters.
@@ -72,9 +72,14 @@ def filter_signal(sig, fs, pass_type, f_range, filter_type='fir',
     ...                          filter_type='fir', f_range=(1, 25))
     """
 
-    check_param_options(filter_type, 'filter_type', ['fir', 'iir'])
+    if filter_type is not None:
+        check_param_options(filter_type, 'filter_type', ['fir', 'iir'])
+    else:
+        # Infer IIR if relevant parameter set, otherwise, assume FIR
+        filter_type = 'iir' if butterworth_order is not None else 'fir'
 
     if filter_type.lower() == 'fir':
+        _fir_checks(butterworth_order)
         return filter_signal_fir(sig, fs, pass_type, f_range, n_cycles, n_seconds,
                                  remove_edges, print_transitions,
                                  plot_properties, return_filter)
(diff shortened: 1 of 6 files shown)


ryanhammonds (approved):

TomDonoghue:
Forgot to tag people who might be interested in this update: @QuirinevEngen, @mwprestonjr, & @voytek - does this make sense / look good in terms of an update related to the filter type discussion that started on slack and described in #321

voytek:
This looks totally reasonable to me. If Q and MJ don't seen any issue, then please merge.

TomDonoghue:
@QuirinevEngen & @mwprestonjr - just following up here on whether you have any thoughts / suggestions on this suggested update about filter specification? if it looks good to y'all, I think we can merge this in

mwprestonjr (commented):
Hey @TomDonoghue, a couple small suggestions relating to 'n_cycles': 

- I like the addition of _fir_checks(); one idea would be to add an additional warning if both 'n_cycles' and 'n_seconds' are set, to explicitly warn about the overwriting behavior. 
- 'n_cycles' is only applicable to the FIR filter so _iir_checks() could raise a ValueError if this is not None, similar to the current behavior for 'n_seconds'.

TomDonoghue:
Hey @mwprestonjr - thanks for checking in, and for the suggestions!

I do like both suggestions, but they are both complicated by the current default definition of `n_cycles` as 3 (in `filter_signal` and in `filter_signal_fir`). Because this makes `n_cycles` always default to 3, it can't be checked against None, nor obviously inferred if the user has changed it (we could check if the value is _not_ 3, but that's not super ideal). 

Because of this, I think the main options for better management of `n_cycles` would be:
- drop the default to 3, make `n_cycles=None` as default, so the user always has to set it, and then we can check as you suggest
    - I'm generally fine with this, but it is a change in behavior. I think we defaulted to 3 because it's a _reasonable_ default (for people who otherwise wouldn't know what to put)
- try and add some checking, with the current default, but it risks being pretty messy and under-determined
- not add any checking for `n_cycles`

ryanhammonds:
It does feel odd that n_cycles defaults to 3, but n_seconds can be passed to implicitly overwrite. Defaulting to n_cyles=None and then defaulting to n_cycles=3 within the func seems logical without breaking much.

The mixing of these fir and iir kwargs seems tricky. Next time a breaking release is made, it may be worth considering something like:

```python
def filter_signal(sig, fs, pass_type, f_range, fir_params=None, iir_params=None,
                  print_transitions=False, plot_properties=False, return_filter=False):
```
This would explicitly separate the two sets of kwargs and maybe simplify the checks a bit, but comes at the cost of breaking things.

TomDonoghue:
Yeh, I totally agree with Ryan - and missed this option in thinking through this before: by defaulting n_cycles to None in the function call, but checking and setting it to 3 if nothing else is defined lets us do the checks we want, without changing any current behavior. I pushed a commit to do this! Note that we do now check that not both `n_cycles` and `n_seconds` are defined (and fail if so), but this is done in the FIR code, not the checks related to `filter_signal`, since it is related to calling `filter_signal_fir` directly as well.

TomDonoghue:
I was thinking about @ryanhammonds other point about separating the arguments in separate dictionaries, and I agree - but I actually think we can do a version of this without it (really) being breaking. In thinking of this, I was thinking that `filter_signal` should really match `compute_spectrum` - they are both "wrapper" function providing access to multiple version of the thing they do. In this case, we don't need `fir_params` & `iir_params`, necessarily, but can have `filter_params`, where the contents should match the filter type (which is what `compute_spectrum` does with `**kwargs`. Even better, we can update to use `filter_params` right now, without really changing how anything functions. 

In the latest commit, I dropped all arguments in `filter_signal` that are fitler type specific, and added `**filter_kwargs`. Doing so leads to changing the `_filtertype_checks` functions into a more general `_filter_input_checks` which checks that the provided arguments are all consistent with the specified (or inferred) filter type).

This all works in a way that is _almost_ non-breaking. Notably, I didn't have to change any tests to work, and because of how the inputs / defaults are managed in the sub-functions, none of the default values for non-explicitly passed in arguments should change. 

The _almost_ part of the non-breaking part is that the API for `filter_signal` does necessarily change a bit - it stiff accepts all the same inputs, but the order switches a bit, so there are things that would work before, that will now error - but because of the order of the previous version, I actually think they are constructions that are pretty unlikely (it's only function calls with a large numbers on unnamed values passed in).

For example, the following would have previously given valid FIR filter with n_cycles of 5:
`filter_signal(sig, fs, 'bandpass', (8, 12), 5)`
This would now fail, because the last input gets processed as filter type instead of n_cycles.

For IIR, it's even more tenious - this is the simplest inputs that would previously work but now fail:
`filter_signal(sig, fs, 'bandpass', (8, 12), 3, None, True, 7)`

I don't think I ever use `filter_signal` in a way that would actually be broken by this change, and I would tend to think it's not very likely many people do (basically because the previous API had a lot of parameters in a not super obvious order, so it would be more likely to use keyword arguments instead of inputs by order for anything after the first ~3 inputs, I think. 

So - what do we think of the update organization?
Sidenote: if we like it, it's not quite finished - the docs might need cleaning up a bit to render nicely, and we should probably add a test for `_filter_input_checks`.

If we don't like this - I'll revert the last commit, and we can go back to keeping all the other updates in the PR

ryanhammonds:
Looks good! I wouldn’t worry too much about about the almost not breaking part - I’d argue named kwargs should be used instead of inferring args as kwargs based on order.

I like the kwarg checking too - my one complaint with **kwargs is accidentally misspelling a kwarg and the func silently accepting. But since there are checks on them, this avoids the issue.

PR #323, opened 2023-09-27
[MNT] - Update and more consistent use of `multidim`

TomDonoghue:
While our functions are by convention defined to operate on 1d signals, we have a `multidim` decorator that is used to allow for applying a specified function across rows of a 2d array, allowing for 2d support. 

Previously, however, the `multidim` function checked and dealt with 1D or 2D inputs, without actually doing anything for 3 or more D, such that a 3D input could be passed through, leading to potentially weird / confusing errors from the downstream function. 

Here:
- update multidim to explicitly fail out if input is > 2D
- update function docs where multidim is used, to note the 1D or 2D support 

@QuirinevEngen - this is related to what you were seeing with the filter function (discussion on slack) - if you want to have a quick look and check here.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -19,7 +19,7 @@ jobs:
       MODULE_NAME: neurodsp
     strategy:
       matrix:
-        python-version: ["3.6", "3.7", "3.8", "3.9", "3.10", "3.11"]
+        python-version: ["3.6", "3.7", "3.8", "3.9", "3.10", "3.11", "3.12"]
 
     steps:
     - uses: actions/checkout@v3
--- a/CITATION.cff
+++ b/CITATION.cff
@@ -1,24 +1,46 @@
 cff-version: 1.2.0
-message: "If you use this software, please cite it as below."
+message: >-
+  If you use this software, please cite it using the metadata from this file.
+type: software
+title: 'neurodsp: neuro digital signal processing'
+authors:
+- given-names: 'Thomas'
+  family-names: 'Donoghue'
+  orcid: 'https://orcid.org/0000-0001-5911-0472'
+- given-names: 'Ryan'
+  family-names: 'Hammonds'
+  orcid: 'https://orcid.org/0000-0002-5032-3241'
+- given-names: 'Scott'
+  family-names: 'Cole'
+  orcid: 'https://orcid.org/0000-0002-6168-9951'
+- given-names: 'Richard'
+  family-names: 'Gao'
+  orcid: 'https://orcid.org/0000-0001-5916-6433'
+- given-names: 'Bradley'
+  family-names: 'Voytek'
+  orcid: 'https://orcid.org/0000-0003-1640-2525'
+repository-code: 'https://github.com/neurodsp-tools/neurodsp'
+url: 'https://neurodsp-tools.github.io/'
+license: Apache-2.0
 preferred-citation:
   type: article
   authors:
-  - family-names: "Cole"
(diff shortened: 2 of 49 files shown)


ryanhammonds:
Do you think it could be useful to allow multidim to reshape a ndim array to 2d? I think I wrote something in a diff project that tracks initial shape e.g. `(2, 3, n)` reshapes to `(6, n)`, computes the function on the 2d array, then reshapes the results back to original dim `(2, 3, m)`. I'm not sure if this would add too much overhead or other implications.

TomDonoghue:
@ryanhammonds - yeh, I think extending `multidim` to support beyond 2D makes sense. Current main branch basically silently does the wrong thing with a > 2D, and my first update here is basically the lazy option of adding the catch & error, but you're totally right that it should be fairly feasible to track the shape of the input array, flatten, and then reshape the outputs. The only thing I'm not totally sure about is whether this would easily generalize to potentially different shaped outputs we get from the different functions we use multidim on - but that should be relatively easy to check. 

Do you want to try out an update for this?

ryanhammonds:
@TomDonoghue I pushed an update that adds support for nd arrays. The decorator now reshapes 3+d to 2d, computes the func, then reshapes the output back to the original 3+d shape.

Multiprocessing could be optionally added to the decorator under the 2d case to speed up computation with an `n_jobs` kwarg. This would extend to any nd case and could make things a faster for functions that don't already multiprocess the 2d case. I could push an update with this too if you think it's worth it.

TomDonoghue:
@ryanhammonds - apologies for dropping off here!

In terms of the multidim decorator specifically: I think this looks good to go in terms of the code? We should just add an test, I think, to have a smoke test at least that passing through the 3D array works, and gets back to the appropriate shape, etc. Do you want to add that? I don't mind doing it if you prefer!

In terms of adding multiprocessing support: that would be great! I don't think it needs to go into this PR really though, so maybe let's merge this when it's ready, and then if you have a chance and want to test out multiprocessing, you can open a new PR for that?

TomDonoghue:
@ryanhammonds this looks great! 

I think it's about ready to merge - just two quick check ins:
- Now that function generally accept arbitrary shaped arrays, I noticed we had a mix of listing this in the docstring as `nd array` or just `array`. I consolidated to `array` (this seems to be what MNE does, and I don't see that adding `nd` really add much info). Does this sound good to you? Happy to revert / consolidate to `nd array` instead if you prefer
- We previously didn't have the multidim decorator on 'pass through' funcs like `filter_signal` or `compute_fluctuations`, because each of the sub-functions that the data gets passed through does have the decorator, so then we distribute there (eg. filter_signal when passed a 3D signal as FIR would get split and looped at the `filter_signal_fir` stage). I'm not sure if I've missed something about the changes here that changes this - do we need to add the decorator to these "wrapper" function s now (it doesn't change much, other then adding them to the wrapper means that each sub-array would have the filter type (or analogous) checked each time

ryanhammonds:
@TomDonoghue for the second point and funcs like filter_signal:

There is now a pass_2d_input kwarg on the multidim decorator. This keeps the same functionality when passing 2d array (it passes through directly to the function). But with 3+d arrays and pass_2d_input=True, the decorator now reshapes to 2d, passes that to the func (e.g. filter_signal), and the reshapes the output back to the original 3+d shape.

TomDonoghue:
Catching up on code things - thanks (belatedly) for the update on this @ryanhammonds!

As far as I know everything here is good to go, so I'm going to merge in now!

PR #327, opened 2024-02-24
[MNT] - Update procedures for "wrapper" funcs

TomDonoghue:
Following from #322, in which `compute_spectrum` was used as a model for updating the 'wrapper' filter function, this PR takes the general approach that was expanded there and applies it back across other examples of using 'wrapper' functions in the module

Updates:
- Update `compute_spectrum` to have a more consistent API and use kwargs input checking
- Add a new 'wrapper' function for instantaneous measures

### DIFF
--- a/doc/api.rst
+++ b/doc/api.rst
@@ -82,6 +82,7 @@ Hilbert Methods
 .. autosummary::
     :toctree: generated/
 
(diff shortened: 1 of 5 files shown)


ryanhammonds (approved):
Looks good!

PR #329, opened 2024-04-14
[ENH] - Add functionality for managing sim params and simulating multiple signals together

TomDonoghue:
In the [Aperiodic Method project](https://github.com/AperiodicMethods/AperiodicMethods), I was simulating large groups of signals while both systematically iterating across parameter values and/or sampling from parameter distributions. In order to do so, I had to develop some utilities for managing simulation parameters and simulating groups of time series together. 

This PR suggests adding that functionality directly into NDSP. It is all quite general, and I think fits well within the sim module. The code added here is basically the code that is demo'd in the aperiodic methods project here (which might be a good starting point to get a feel for how this approach looks / works): https://aperiodicmethods.github.io/notebooks/12-TimeSimsParameters.html

I've also sketched out some tutorials, which give a broad overview of the general approach / how this code works.


neurodsp/sim/multi.py
```diff
@@ -0,0 +1,166 @@
...
+        If 'array', all sims are organized into a 2D array.
+
+    Returns
+    -------
+    sims : dict of {float : array} or array
+        If dict, dictionary of simulated signals, where:
+            Each key is the simulation parameter value for the set of simulations.
+            Each value is the set of simulations for that value, as [n_sims, sig_length].
+        If array, is all signals collected together as [n_sims, sig_length].
+    """
```
ryanhammonds, line 119: Having a float as a key feels odd. Maybe a list of [{param: float, sig: array}, ...]? 

Or a 1d array of float (params) and a 2d array (sigs), separately returned? And then replace `output` kwarg with `return_params=bool`. -> just saw this is done in `sim_sampler` so maybe that would work here?

neurodsp/sim/params.py
```diff
@@ -0,0 +1,589 @@
...
+    def __getitem__(self, label):
+        """Make object subscriptable, to access stored simulation parameters.
+
+        Parameters
+        ----------
+        label : str
+            Label to access simulation parameters from `params`.
+        """
+
+        return {**self.base, **self._params[label]}
```
ryanhammonds, line 50: If user has to initialize object with n_seconds and fs, why is this return in base? Is it so the dict can be quickly unpacked into a sim func?

neurodsp/sim/params.py
```diff
@@ -0,0 +1,589 @@
...
+class SimIters(SimParams):
+    """Object for managing simulation iterators.
+
+    Parameters
+    ----------
+    n_seconds : float
+        Simulation time, in seconds.
+    fs : float
+        Sampling rate of simulated signal, in Hz.
+    sim_params : SimParams
```
ryanhammonds, line 258: It feels odd that this class both inherits from SimParams and accepts a SimParams as initialization argument. It may be worth considering to choose one way or the other if possible.

A hack may be to add a method to SimParams that returns an initialized SimIters object. And then drop sim_params arg here.

neurodsp/sim/params.py
```diff
@@ -0,0 +1,589 @@
...
+    Parameters
+    ----------
+    n_seconds : float
+        Simulation time, in seconds.
+    fs : float
+        Sampling rate of simulated signal, in Hz.
+    n_samples : int, optional
+        The number of parameter iterations to set as max.
+        If None, samplers are created as infinite generators.
+    sim_params : SimParams
```
ryanhammonds, line 438: Same comment as left on SimIters about having a method on SimParams that returns an initialized SimSampler, e.g.:

```
sim_sampler = sim_params.to_sampler(n_samples)
sim_iters = sim_params.to_iter()
```

This may reduce complexity of the either/or in the init here.

neurodsp/sim/update.py
```diff
@@ -0,0 +1,360 @@
...
+
+    return lambda params, value : params['components'][component].update({parameter : value})
+
+
+def create_updater(update, component=None):
+    """Create an updater function for updating simulation parameters.
+
+    Parameters
+    ----------
+    parameter : str
```
ryanhammonds, line 56: name of this arg -> `update`

tutorials/sim/plot_SimMulti.py
```diff
@@ -0,0 +1,120 @@
...
+# from the same set of parameters.
+#
+
+###################################################################################################
+
+# Define a set of simulation parameters
+params = {'n_seconds' : 5, 'fs' : 250, 'exponent' : -1, 'f_range' : [0.5, None]}
+
+# Simulate multiple iterations from the same parameter definition
+sigs = sim_multiple(sim_powerlaw, params, 3)
```
ryanhammonds, line 28: Can `n_seconds` and `fs` can be varied? It seems not since `sim_multiple` returns an array, but maybe it can be a jagged array. If not though, maybe `sim_multiple(n_seconds, fs, ...)` instead

tutorials/sim/plot_SimMulti.py
```diff
@@ -0,0 +1,120 @@
...
+
+###################################################################################################
+# SigIter
+# -------
+#
+# In some cases, it may be useful to define a way to sample iterations from the same set of
+# simulation parameters. To do so, we can use the :class:`~.SigIter` class.
+#
+# Using this class, we can define an object that stores the simulation function, the
+# set of parameters, and optional a number of simulations to create, and use this object
```
ryanhammonds, line 43: optional -> optionally?

ryanhammonds (commented):
Looks good! I reviewed all the module code. I probably should've started with the tutorials to get a better sense of usage, so sorry if some of the comments lack foresight.

The tests are the only thing I didn't review. I can come back to those if you want. But since they pass / have coverage, I'm happy with them.

neurodsp/sim/multi.py
```diff
@@ -0,0 +1,166 @@
...
+    sim_params : dict
+        The parameters for the simulated signal, passed into `sim_func`.
+    n_sims : int
+        Number of simulations to create.
+
+    Returns
+    -------
+    sigs : 2d array
+        Simulations, as [n_sims, sig length].
+    """
```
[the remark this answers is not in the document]
TomDonoghue: Yes, good idea! I've added docstring examples to the public facing functions in `multi`

neurodsp/sim/update.py
TomDonoghue: good catch - fixed!

neurodsp/sim/multi.py
```diff
@@ -0,0 +1,166 @@
...
+    ----------
+    sim_func : callable
+        Function to create the simulated time series.
+    sim_params : iterable
+        The parameters for the simulated signal, passed into `sim_func`.
+    return_sim_params : bool, optional, default: False
+        Whether to yield the simulation parameters as well as the simulated time series.
+    n_sims : int, optional
+        Number of simulations to set as the max.
+        If None, creates an infinite generator.
```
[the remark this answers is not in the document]
TomDonoghue: I think in most cases one would know the number of simulations you want to use for a given application, but in at least some cases, the user may not know up front, and sample indefinitely until a stopping criterion is hit (for example, simulate samples to compute a value of interest, and do not stop until the distribution of properties has some value)

neurodsp/sim/multi.py
```diff
@@ -0,0 +1,166 @@
...
+    ------
+    sig : 1d array
+        Simulated time series.
+    sample_params : dict
+        Simulation parameters for the yielded time series.
+    """
+
+    if len(sim_params) and n_sims and n_sims > len(sim_params):
+        msg = 'Cannot simulate the requested number of sims with the given parameters.'
+        raise ValueError(msg)
```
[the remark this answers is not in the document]
TomDonoghue: Yeh - `sim_params` is a flexible input of an iterable, so it can be finite, in which case it effectively defines the number of sims, and `n_sims` is optional, and can define another way to define the number of sims (for example, if `sim_params` is infinite). 

Examples that show the different possibilities here:
```
from itertools import repeat
from neurodsp.sim.multi import sig_sampler
from neurodsp.sim.aperiodic import sim_powerlaw

params = [{'n_seconds' : 5, 'fs' : 250, 'exponent' : -1},
          {'n_seconds' : 5, 'fs' : 250, 'exponent' : -2}]
# This works - creates a generate for 2 signals
gen = sig_sampler(sim_powerlaw, params)

# This fails - n_sims requested beyond the length of params
gen = sig_sampler(sim_powerlaw, params, n_sims=3)

# This works - uses n_sims to define the length of infinite params def
gen = sig_sampler(sim_powerlaw, repeat(params[0]), n_sims=3)
```

I don't think this was super obvious from the docs / code (especially that the number of sims could be bound by the length of `sim_params` if `n_sims` is not set), so I've updated that a bit here (also to more properly support iterable inputs). More generally, this function isn't a user facing function - but the same logic applies to users passing things in through `sim_from_sampler`).

neurodsp/sim/params.py
TomDonoghue: Yeh,  "base" is just a shorthand to get the common parameters of n_seconds & fs (that are defined across all functions), and return with the specific parameters of some definition. Base parameters are included in order to pass this all into a sim funtion

tutorials/sim/plot_SimMulti.py
TomDonoghue: Within the context of `sim_multiple` the `n_seconds` and `fs` parameters cannot be varied (they are fixed) - though in this function so are all other simulation parameters - this function is just a shortcut to simulate multiple samples of a given function + parameter set. 

More broadly, varying `n_seconds` & `fs` was not part of my use case and is therefore a bit of a blind spot for this PR I think - as you note it adds the complexity that the size / shape changes. I might come back to this and see if there is a useful extension to support it.

neurodsp/sim/params.py
TomDonoghue: Good catch / note - it works really nicely to have these "export" methods defined on SimParams that one can use if you already have a SimParams object, and then the derived objects (SimIters, SimSamplers) no longer optionally accept a SimParams object at init, and have a cleaner initialization! Updated!

neurodsp/sim/params.py
TomDonoghue: Yes - agreed! Updated (notes in next comment)

neurodsp/sim/combined.py
```diff
@@ -161,6 +161,38 @@ def sim_peak_oscillation(sig_ap, fs, freq, bw, height):
...
+
+    Returns
+    -------
+    sig : 1d array
+        Simulated combined peak signal.
+    """
+
+    sim_names = list(components.keys())
+    assert len(sim_names) == 2, 'Expected only 2 components.'
+    assert sim_names[1] == 'sim_peak_oscillation', \
```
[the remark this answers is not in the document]
TomDonoghue: Yeh, this is a good point - I think the slight tension is whether `sim_combined_peak` should mimic the API of `sim_combined` (as current, but then it has extra constraints) or mimic `sim_peak_oscillation`. I think you're right - the usage here is a bit more bespoke, and so accepting a very general dict input but requiring specific things doesn't really make sense (as opposed to updating the API to list the expected things). 

I'll update!

neurodsp/sim/multi.py
```diff
@@ -0,0 +1,166 @@
...
+    sims : dict of {float : array} or array
+        If dict, dictionary of simulated signals, where:
+            Each key is the simulation parameter value for the set of simulations.
+            Each value is the set of simulations for that value, as [n_sims, sig_length].
+        If array, is all signals collected together as [n_sims, sig_length].
+    """
+
+    sims = {}
+    for ind, cur_sim_params in enumerate(sim_params):
+        label = sim_params.values[ind] if hasattr(sim_params, 'values') else ind
```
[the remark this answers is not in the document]
TomDonoghue: Yeh, good call - this reflects that it accepts the new `ParamIter` object, which was not documented. Fixed!

neurodsp/sim/multi.py
TomDonoghue: Yeh, so the more I thought about this, and trying to think about something like a structured dictionary of the kind you mentioned - at some point I realized that really that just becomes an object. So, having fallen down that rabbit hole, I've updated this PR to add a couple basic objects to manage groups of simulations (and related metadata, like parameters).  The sim multi functions now default to returning these objects (though it can be specified to return arrays, with no metadata). I think this is probably the best way to go here - everything is now updated

neurodsp/sim/combined.py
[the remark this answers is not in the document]
TomDonoghue: Edit: I later realized why I did it like this - mimicking sim_combined allows for writing out groups of simulations in the same way, and then passing them through the either `sim_combined` or `sim_combined_peak` in the same way (even though `sim_combined_peak` has some extra requirements), and I think this is useful - changing this would reduce this consistency

TomDonoghue:
I think everything pertaining specifically to the updates here is addressed, so I'm going to merge this in now to move forward with integrating this new approach into a project. We should test this whole approach out a bit more before tagging the next version release.

PR #334, opened 2024-08-12
[BUG] burst_def bug

ryanhammonds:
The docstring of sim_bursty_oscillation says a 1d array can be passed to the burst_def kwarg. Passing a 1d array leads to an error here, since `burst_def=='prob'` returns an array:

```python
if burst_def == 'prob' and burst_param not in burst_params:
```

This PR restructures the function with appropriate conditionals to fix. If 1d arrays aren't meant to be supported, then we'd just need to update the docstring.


TomDonoghue:
Hey @ryanhammonds I merged this to main to get the tests running, and the tests are failing. 

Based on looking through what's failing is things like this:
```
sig = sim_bursty_oscillation(n_seconds=10, fs=500, freq=5, enter_burst= 0.2, leave_burst=0.8)
```
There's a bit of sloppiness here (I think this comes from the old API) since the 'enter_burst' and 'leave_burst' parameters should arguably be packaged into a 'burst_params' input dict (which would work), but this is done in places here as [here](https://github.com/neurodsp-tools/neurodsp/blob/1e7959a9dacaddae375ac5f8039568bf7617122c/neurodsp/plts/rhythm.py#L30), and if we don't want to make any breaking changes, we probably want to maintain this. 

The test case we need to work is something like this, right (this should have been in the tests really):
```
burst_def = np.array([True, False, True, False, True, False, True, False, True, False])
sim_bursty_oscillation(1, 100, 10, burst_def=burst_def)
```

The failure point, as you found, is: `if burst_def == 'prob' and burst_param not in burst_params:`, from 
[here](https://github.com/neurodsp-tools/neurodsp/blob/1e7959a9dacaddae375ac5f8039568bf7617122c/neurodsp/sim/periodic.py#L157), which fails at type checking `burst_def` in the case it's an array (this used to work, but type checking against arrays has become stricter).

As an alternative fix here, I think we can just update this type check to the following:
`if isinstance(burst_def, str) and burst_def == 'prob' and burst_param not in burst_params:`

Does that make sense? I'm pretty sure that addresses the issue here, and in not moving around any ordering, preserves all previous functionality. 

If this make sense to you, I think we can replace this PR by updates that:
- update this type-checking line to the version above that tests for being a string
- add an explicit test of burst_def as an array input (for example, from above)
- sweep through and update examples that pass burst_params in the old style, so we model this approach better

Does this make sense / sound good to you? If so, I can PR this

ryanhammonds:
@TomDonoghue That sounds good to me. You're proposed fix looks clean / backward compatible. Thanks for look into this!

TomDonoghue:
@ryanhammonds - okay sounds good - I added these changes in #349 - if that looks good, we can merge there, and close this! Thanks for finding / figuring out this bug!

TomDonoghue:
Updates done in #349

PR #339, opened 2024-11-16
[MNT] - Fix power spectrum plot ylabel formatting

TomDonoghue:
Currently, the default y-label for a PSD plot is the string literal 'Power (V^2/Hz)', which isn't properly formatted to display the superscript exponent. 

This PR updates the formatting, to actually use a superscript, so the plot label now looks like this:
Screen Shot 2024-11-16 at 4 37 52 PM

I think the only thing to consider / check is whether there is some reason this formatting might not work across systems(?). It works smoothly for me - would be good if someone else could check it renders properly.

### DIFF
--- a/neurodsp/plts/spectral.py
+++ b/neurodsp/plts/spectral.py
@@ -61,7 +61,7 @@ def plot_power_spectra(freqs, powers, labels=None, colors=None, ax=None, **kwarg
         ax.loglog(freq, power, color=color, label=label)
 
     ax.set_xlabel('Frequency (Hz)')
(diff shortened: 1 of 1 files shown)


ryanhammonds (approved):
Seems like [mathtext](https://matplotlib.org/stable/users/explain/text/mathtext.html) is distributed with matplotlib:

> TeX does not need to be installed to use Mathtext because Matplotlib ships with the Mathtext parser and engine.

My understanding is this should work with any installation of matplotlib and doesn't depend on a system latex installation.

PR #345, opened 2024-12-20
Replace np.NaN with np.nan for NumPy 2.x

penguinpee:
The deprecated `np.NaN` has been removed from NumPy 2.x


penguinpee:
Nevermind. I forgot, I had already submitted that. And it has been merged (PR #333).

Issue #312, opened 2023-06-02
Add description in the data folder


TomDonoghue:
The readme in `data/` should have a bit more description - including notes on fs, data length, filter range that was applied to the filtered example, and also a brief note on where this data comes from.

TomDonoghue:
Done in #350

PR #350, opened 2025-02-11
[MNT] - Update example data file info & usage

TomDonoghue:
Responds to #312 

I did a quick check through for information on the data files we use, and added a description to the README file in the data folder to add the basic information there. Notably, it adds a brief description of what the data is, sampling rate, and related publication. Having moved this info, I removed a bit of info in the tutorials themselves. 

Also, we have a pre-computed filtered version of the DBS data file, and I don't remember why. It's barely used in the tutorials, and where it was (instantaneous measures tutorial) there's not really any reason to, so I updated this tutorial to simply compute the filtered version. This means that `data_sample_1_filt` is now not used anymore. I don't think we should remove it, at least for now (if we did, anyone running an older example with `load_ndsp_data` that uses it would get an error, which I don't think helps anything - so I think we can consider the file deprecated, but leave it there for now.

### DIFF
--- a/data/README.txt
+++ b/data/README.txt
@@ -2,3 +2,14 @@ Data
 ====
 
 Example and test data files for NeuroDSP.
+
+The following data files are made available with the module, for use in the tutorials:
+- `sample_data_1`: a segment of human primary motor cortex (M1)
+    - Data sample is from a DBS device implanted in a patient with Parkinson's Disease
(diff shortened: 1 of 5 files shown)


ryanhammonds (approved):
Looks good. The only thing I noticed is that `sample_data_1_filt.npy` is no longer used anywhere, I think. So should it be dropped from `data/`?

Edit: nevermind! just re-read your original comment

PR #353, opened 2025-03-21
Update dualthresh.py

kguarian:
cast arrays to fix type error

### DIFF
--- a/neurodsp/burst/dualthresh.py
+++ b/neurodsp/burst/dualthresh.py
@@ -169,6 +169,8 @@ def _rmv_short_periods(sig, n_samples):
     if osc_ends[-1] < osc_starts[-1]:
         osc_ends = np.append(osc_ends, len(osc_changes))
 
+    osc_starts, osc_ends = np.array(osc_starts), np.array(osc_ends)
+    
     osc_length = osc_ends - osc_starts
     osc_starts_long = osc_starts[osc_length >= n_samples]
     osc_ends_long = osc_ends[osc_length >= n_samples]


neurodsp/burst/dualthresh.py
```diff
@@ -169,6 +169,8 @@ def _rmv_short_periods(sig, n_samples):
     if osc_ends[-1] < osc_starts[-1]:
         osc_ends = np.append(osc_ends, len(osc_changes))
 
+    osc_starts, osc_ends = np.array(osc_starts), np.array(osc_ends)
```
ryanhammonds, line 172: Should casts should be on 163 and 165? e.g.
```
osc_starts = np.array([0])
osc_ends = np.array([len(osc_changes)])
```
Outside of 163 and 16, osc_starts and osc_ends should be np.array already I think. Casting array to array doesn't make sense / may have unnecessary overhead.

ryanhammonds (commented):
Thanks!

Do you have a minimal reproducible example that raises the error? Will be useful to get an idea of when the errors occurs.

neurodsp/burst/dualthresh.py
kguarian: I wholly agree. It doesn't make too much sense, but I was getting an error where the line, "osc_length = osc_ends - osc_starts" gave an error saying subtraction is undefined between lists. This suggests that osc_ends and osc_starts were lists. I added the line then dualthresh ran without error.

ryanhammonds: Thanks, I understand the error. My question was: should the casts occur where the lists are created, on lines 163 and 165, instead of where you are casting them in line 175?

Do you have a minimal reproducible example you could add to the PR, e.g. a small simulation that causes the error? I want to poke around it a bit to get a better sense of what's going on.

kguarian:
https://github.com/kguarian/publish_the_paper/tree/main/sim_signal_kit

This module reproduces the error. You can't miss it.

PR #354, opened 2025-10-12
[ENH] - Extend spectral power with more FFT options

TomDonoghue:
Add a function to compute a spectrum as from the FFT directly.

Why: every so often I find myself wanting to compute a single FFT window, and end up messing around remember how to do so. This adds the option to compute a spectrum from an individual FFT window, following the neurodsp spectral API. 

Note that this PR is already merged with #358. In adding this function, since it uses numpy (not scipy like the others) there is no a padding option already available. I added a simplified padding function for a single window, adapted from the one we had before and that is suggested to be removed in #358. 

Also, at one point I was using `get_positive_fft_outputs` to select positive ffts, but then realized this can be short-cutted by using `np.rfft` instead of `np.fft`. As such, I think the `get_positive_fft_outputs` can actually be removed, unless anyone thinks it's useful for some reason?

### DIFF
--- a/neurodsp/spectral/utils.py
+++ b/neurodsp/spectral/utils.py
@@ -130,43 +130,6 @@ def trim_spectrogram(freqs, times, spg, f_range=None, t_range=None):
     return freqs_ext, times_ext, spg_ext
 
 
-def get_positive_fft_outputs(freqs, powers=None, drop_zero=False):
-    """Get the positive frequency values for an FFT.
-
-    Parameters
-    ----------
-    freqs : 1d array
-        Frequency vector corresponding to the FFT estimate, with positive & negative frequencies.
-    powers : 1d array, optional
-        Complex power value estimates from the FFT.
-    drop_zero : bool, optional, default: False
-        Whether to drop the estimate for frequency of 0.
-
-    Returns
-    -------
-    freqs : 1d array
-        Frequencies at which the measure was calculated.
-    spectrum : array
-        Power spectral density.
-        Only returned if an input power spectrum is passed.
-
-    Notes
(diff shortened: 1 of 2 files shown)


neurodsp/spectral/power.py
```diff
@@ -136,6 +139,50 @@ def compute_spectrum_wavelet(sig, fs, freqs, avg_type='mean', **kwargs):
     return freqs, spectrum
 
 
+@multidim(select=[0])
+def compute_spectrum_fft(sig, fs, window=None, nfft=None, f_range=None):
```
ryanhammonds: could be added to \_\_init__.py so it can be directly from neurodsp.spectral like the other compute_spectrum_* functions

neurodsp/spectral/power.py
```diff
@@ -136,6 +139,50 @@ def compute_spectrum_wavelet(sig, fs, freqs, avg_type='mean', **kwargs):
...
+    spectrum = np.abs(np.fft.rfft(sig)) ** 2.
+    freqs = np.fft.rfftfreq(len(sig), 1. / fs)
+
+    if f_range:
+        freqs, spectrum = trim_spectrum(freqs, spectrum, f_range)
+
+    return freqs, spectrum
+
+
 def compute_spectrum_welch(sig, fs, avg_type='mean', window='hann', nperseg=None,
```
ryanhammonds: Was the plan to convert this implementation to scipy's implementation?

neurodsp/spectral/power.py
TomDonoghue: Yeh, good call - added!

PR #356, opened 2025-10-19
[ENH] - Add signal yielder functionality

TomDonoghue:
This PR add some helper / utility functions for iterating across arrays. With long time series, in some cases it can be useful to iterate across segments. It adds a function to step across time (for a 1D or 2D array) and a separate function for stepping across 'channels' (across rows within a 2D array).

### DIFF
--- a/doc/api.rst
+++ b/doc/api.rst
@@ -525,3 +525,13 @@ Outliers
     remove_nans
     restore_nans
     discard_outliers
+
+Yielders
+~~~~~~~~
+
+.. currentmodule:: neurodsp.utils.yielders
+.. autosummary::
+    :toctree: generated/
+
+    step_over_time
+    step_over_signals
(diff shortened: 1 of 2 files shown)


ryanhammonds (approved):
Looks good! The I see the usefulness of step_over_time. I'm unsure the use case for:
```python
for sig in step_over_signals(signals):
    pass
```
This seems more verbose than calling:
```python
for sig in signals:
    pass
```
Maybe the generator is useful for other cases:
```
gen = (sig for sig in signals)
next(gen)

gen = step_over_signals(signals)
next(gen)
```

Anyways this looks good to me!

TomDonoghue:
Thanks for checking @ryanhammonds! 

I agree the time stepper is more useful, I added the channel stepper somewhat for completeness. The use case I was thinking of like the last version you noted - creating a generator at some point, and calling next as needed (I've been doing things like this in the use case that I've been using these in, which is making some gifs).

Issue #357, opened 2025-10-19
Welch's method: padding & code support


TomDonoghue:
Short version: I think we should update to use `scipy.signal.welch` including using it's options for padding. 

Longer version:
In `compute_spectrum_welch` we use `scipy.signal.spectrogram` to do most of the work. As a brief historical note, we use this because even though there is also Welch's function in scipy (https://docs.scipy.org/doc/scipy/reference/generated/scipy.signal.welch.html), back in the day we were quite interested in median Welch, and you couldn't originally do this in in their implementation - so we used spectrogram to get all the windows and did our own averaging. As of scipy 1.2.0, the welch's function has an average function, so this is no longer relevant. 

As pointed out in #344 - `scipy.signal.spectrogram` is marked as legacy, so we should move away from it. There is a new spectrogram method in scipy we could use, but I think for our current purposes we should just use the available Welch's method. This would add a minimum version of scipy (v1.2.0), but not a strict one (current scipy is on v1.16.X).

What I wanted to check in on though is that we also have our own signal padding (https://github.com/neurodsp-tools/neurodsp/pull/318). I think moving to `scipy.signal.welch` supports what we want, using the `nfft` input to do zero-padding. This is potentially actually better in that something I didn't realize at the time but in #318 I think what happens is we insert zero's so that when the signal gets re-segmented it has the zeros - but the spectrogram function doesn't "know" this, with the issue being that this might mean that applying a window function gets applied incorrectly (gets applied to the already zero-padded window, instead of being applied prior to zero-padding, which I think is what should happen, and presumably what happens if we do padding through scipy's Welch's. I didn't poke around to test this, but the outputs could be compared to check. Unless I'm missing something, I think we can replace our own padding with what scipy does, and arguably this would be slightly better.

@ryanhammonds - I'm not sure if I'm forgetting and/or misinterpreting something about #318 where the padding options were added. If we switch to `scipy.signal.welch` is there a reason to maintain our own padding functionality? If not, I think we can update to oursource to scipy while also addressing the deprecations.

TomDonoghue:
Okay sounds good - I think this all makes sense and we're on the same page!

I also did a bit of comparing, to make sure we don't accidentally introduce any big / unexpected changes. It does seem the two approaches are effectively equivalent. 

** Basic Comparison **

If a (simplified) version of the new version is this:
```
    nperseg, noverlap = check_spg_settings(fs, window, nperseg, noverlap)    
[...]

Issue #344, opened 2024-12-18
Using SciPy's ShortTimeFFT instead of spectrogram function for compute_spectrum


noortasnim:
This issue has been somewhat brought up in [another post](https://github.com/neurodsp-tools/neurodsp/issues/325), but I also wanted to point out that the scipy.spectrogram function used in [compute_spectrum_welch](https://github.com/neurodsp-tools/neurodsp/blob/eca4dfd49b1fc3c40320c008a6e6b237be49aa94/neurodsp/spectral/power.py#L219) is now considered a [legacy function](https://docs.scipy.org/doc/scipy/reference/generated/scipy.signal.spectrogram.html).

There's a possibility that it gets deprecated in future SciPy versions. To avoid an issue in the future, would it be possible to use [ShortTimeFFT.spectrogram](https://docs.scipy.org/doc/scipy/reference/generated/scipy.signal.ShortTimeFFT.spectrogram.html#scipy.signal.ShortTimeFFT.spectrogram) instead?

PR #358, opened 2025-10-27
[MNT] - Update Welch implementation

TomDonoghue:
Responds to #357 (and related to #344). See #357 for notes and checks of the new implementation. 

Updates:
- update the implementation of `compute_spectrum_welch` to use `scipy.signal.welch`
- remove our window padding options, as this is now available through `scipy.signal.welch`
- update the checker function, including updating the name (this is not public facing so I don't think should be an issue), and updating for the checker function to support the `fast_len` input
- removes the `outliers_percent` option from `compute_spectrum_welch`, as this is no longer supported using this implementation


ryanhammonds (approved):
looks good to me!

PR #364, opened 2026-09-16
The formula in the docstring of the `sim_knee` function has been revi…

zamanzadeh98:
…sed.

The docstring formula grouped the knee term incorrectly. It now matches the code: knee is added inside the parentheses before multiplying by f**exponent1.


TomDonoghue:
Good catch! Thanks @zamanzadeh98 for the fix - I'll merge this in. 

Also, slight tangent, but we might as well update the docs to use proper latex rendering for formulas for the docsite, so I've opened an issue for this task in #365

